PR #358, opened 2018-02-02
Scopy preferences

DanielGuramulta:
Implementation of the preferences feature for scopy


src/preferences.cpp
```diff
@@ -0,0 +1,165 @@
+#include "preferences.h"
```
AlexandraTrifan: Could you add license info for this file as well?

DanielGuramulta: I've added the license with: a68fe4a9f581b1bdf2b299cb383bb8ae871911ae

adisuciu:
Can you please solve the conflict related to the previous pull request ?

PR #370, opened 2018-02-20
Autoset

adisuciu:
The autoset feature is implemented with this pull request. The autoset will perform FFT on the captured data and try to find the highest amplitude tone in the FFT. This FFT is performed multiple times on different sample rates to provide accuracy. Once the tone is found, the algorithm scans for the min/max of the waveform, and updates the vertical/horizontal controls. Also sets the trigger to (max-min)/2 - at 50%.

### DIFF
--- a/src/trigger_settings.hpp
+++ b/src/trigger_settings.hpp
@@ -88,6 +88,8 @@ namespace adiscope {
 			const QPair<double, double>& range);
 		void setTriggerHystRange(int chn,
 			const QPair<double, double>& range);
+		void setTriggerEnable(bool);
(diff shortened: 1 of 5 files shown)


src/trigger_settings.hpp
```diff
@@ -96,12 +98,14 @@ namespace adiscope {
 		void setAdcRunningState(bool on);
 		void onSpinboxTriggerLevelChanged(double);
 
-	private Q_SLOTS:
+		void on_intern_en_toggled(bool);
```
AlexandraTrifan: I guess we can move this back to private slots, right?

Issue #234, opened 2017-08-23
Adding of cursor/position option


tagoylo:
Scopy build: [018987a2](https://github.com/analogdevicesinc/scopy/commit/018987a29e4cfb9b893146feaf5d9b3ef08619f4)

Adding of cursor/position option to see exact frequency value at -3dB point, also to track phase shift along with the magnitude. This will provide convenience for the user in using the network analyzer. Instead of dragging/moving the mouse cursor along the trace to find the -3dB point of a LPF frequency response, for example. The user will just go to the cursor/position option to locate the point.

adisuciu:
Implemented with https://github.com/analogdevicesinc/scopy/pull/372

amiclaus:
On Scopy build: 6bd7075 (v0.11) the feature is implemented.

PR #372, opened 2018-02-21
Network analyzer cursors

claudiupop:
Implemented feature #234


adisuciu:
Bottom bar has some gradient issues
The cursor settings button doesn't do anything and should be removed.
The cursors from magnitude/phase should be locked together.
Move cursor widget inside plot, so it doesn't obstruct any ruler values.

PR #377, opened 2018-02-22
Manual calibration

mirceacaprioru:
This is a pull request for merging the manual calibration tool. This tool is hidden and manual_calibration property must be enabled inside .ini file. 

Also in order to calibrate the power supply m2k_calib.ini must be created on m2k mass storage drive. 
Once the calibration procedures are executed m2k_calib.ini must be overwritten and m2k ejected in order to write the new calibration values into the context attributes in IIO. 

For the waveform generator and oscilloscope the automatic calibration is still in place and is executed at scopy startup.


adisuciu:
Please fix conflicts and resubmit.

src/manualcalibration.h
```diff
@@ -0,0 +1,153 @@
...
+
+	Calibration *calib;
+
+	struct stCalibStory stCalibrationStory;
+	struct stCalibParam stParameters;
+	QStringList calibListString;
+	QMap<QString, int> calibOption;
+
+	/*Calibrations procedure stories*/
+	const QStringList positiveOffsetStory = (QStringList() <<
```
dNechita: Also, can be static const in the .cpp file. There's no need for a string list, a simple string will do.
Here's a nice way to declare a text block that contains line endings, spaces, tabs, etc.
```
QString myString(R"(
A text block with line endings
   and
    spaces
   and
    tabs
)");
```
More about raw string literal can be found here: http://en.cppreference.com/w/cpp/language/string_literal

src/tool_launcher.hpp
```diff
@@ -244,6 +250,9 @@ class ToolLauncher_API: public ApiObject
 	bool debugger_enabled();
 	void enable_debugger(bool);
 
+	bool manual_calibration_enabled();
```
dNechita: Method can be const.

src/manualcalibration.h
mirceacaprioru: These are QStringList because they hold multiple steps. The stories for these steps must be separated in some way.

dNechita: Ok, it makes sense.

PR #415, opened 2018-04-24
Tool launcher & user notes enhancement

AlexandraTrifan:
This pull request implements two major features: 

1. Modify the tool launcher such that information about each device is presented in its corresponding page. Also, this new approach allows the user to check information about different devices, while being connected to one.

2. Add a new section for user notes which allows a user to load and view .html files. This can be easily enabled/disabled using the Preferences panel.


ui/device.ui
```diff
@@ -92,49 +143,170 @@ background-color: green;
...
+        <property name="sizePolicy">
+         <sizepolicy hsizetype="Minimum" vsizetype="Maximum">
+          <horstretch>0</horstretch>
+          <verstretch>0</verstretch>
+         </sizepolicy>
+        </property>
+        <property name="styleSheet">
+         <string notr="true">QLabel {
+color: white;
+font-family: &quot;monospace&quot;;
```
claudiupop: You should remove the font because it overrides the font set in the app

ui/preferences.ui
```diff
@@ -947,6 +886,192 @@ QCheckBox::indicator:checked { background-color: rgb(74,100,255); }</string>
...
+                 <verstretch>0</verstretch>
+                </sizepolicy>
+               </property>
+               <property name="styleSheet">
+                <string notr="true">QCheckBox {
+  spacing: 8px;
+  background-color: transparent;
+  font-size: 14px;
+  font-weight: bold;
+  font-family: &quot;monospace&quot;;
```
claudiupop: You should remove the font because it overrides the font set in the app

ui/preferences.ui
```diff
@@ -947,6 +886,192 @@ QCheckBox::indicator:checked { background-color: rgb(74,100,255); }</string>
...
+             </property>
+             <item>
+              <widget class="QCheckBox" name="doubleClickCheckBox">
+               <property name="styleSheet">
+                <string notr="true">QCheckBox {
+  spacing: 8px;
+  background-color: transparent;
+  font-size: 14px;
+  font-weight: bold;
+  font-family: &quot;monospace&quot;;
```
claudiupop: You should remove the font because it overrides the font set in the app

ui/device.ui
AlexandraTrifan: Sure, sorry. Should be fixed with commit bc08231.

PR #446, opened 2018-06-11
JavaScript: add script examples

amiclaus:


amiclaus:
Headers + Readme will be included.

For some instruments, I think there are too many attributes that should be included to create a generic setup function.

It's important that the Scopy Reset function to be scriptable (in the near future). A good practice would be to perform a reset to initial values before running the actual script .

rgetz:
Yeah, all_instrument_reset() would be good - since you don't know the state of the hardware when you run the script.

AlexandraTrifan:
I am working on making the Scopy Reset function scriptable.

hajime725:
How to show console ?
where is printToConsole()  output?
I can't find any console.
My environment is windows10 64bit,scopy-f21d58f.

amiclaus:
You need to run the script from a Unix command line to get the output of the printToConsole(). 

To see how to run the script from command line, please check:
https://wiki.analog.com/university/tools/m2k/scopy/scripting-guide

hajime725:
how to use Unix command line in windows10?
I try "scopy --script power_dmm.js" in cmd.exe and PowerShell and bash on Windows.
Environment variable PATH is added.
And SCOPY can  execute from cmd.exe and PowerShell but can't execute in bash on windows.

amiclaus:
You can try using MSYS2 which provides a bash shell. https://www.msys2.org/

Also, if you run the script from the command line, you need to add the connect function:

```
function connect(host) {
	print("Connecting to " + host + "...")

	var success = launcher.connect("ip:" + host)

	if (success)
		print("Connected!")
	else
		print("Failed!")

	return success;
}
```

```
function main() {

	var host = "192.168.2.1"

	var connected = connect(host)
	if (!connected)
                     return Error()

        launcher.disconnect()
}
```

hajime725:
thanks!
run successfully.

PR #448, opened 2018-07-03
Run external commands/scripts and make the manual calibration scriptable (experimental)

AlexandraTrifan:
1) Create the minimal structure which allows the user to execute external shell scripts/commands in Scopy using the existing Scopy Scripting Environment. In order to enable this, some extra steps are required.

2) Add a scriptable API to manipulate some of the manual calibration functionalities using scripts. Similar to the external scripts, in order to enable this, some extra steps are required.


AlexandraTrifan:
I added some commits to handle the manual calibration using scripts and I also updated the description and name of the pull request.

PR #451, opened 2018-07-11
Save detached tools

DanielGuramulta:
Save/ Load the state of a detached tool(position on screen, minimized, maximized)

### DIFF
--- a/src/detachedWindow.cpp
+++ b/src/detachedWindow.cpp
@@ -23,11 +23,13 @@
 
 using namespace adiscope;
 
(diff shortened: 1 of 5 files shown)


rgetz:
This is great - common problems are :

 - save with dual screens, (with content on two screens) and then unplug second screen, and run again. 

 - save with large (4K) screen resolution, with some windows on bottom, then change resolution to less (1080) and then restore. 

These open windows shouldn't get lost, or painted in locations that the user can't see.

rgetz:
Thanks - this has always been one of my pet peeves of poorly designed software.

PR #453, opened 2018-07-12
Print icon

DanielGuramulta:
Small UI changes regarding the print functionality of some tools.


adisuciu:
Doesn't save screenshot when no extension is typed in.

PR #471, opened 2018-08-01
build,CI/travis: fix build on Linux & Mac OS X

commodo:
Well, predominantly on Linux, the build is currently failing, but the scripts don't show that status.

This change-set does a rework to make Scopy build on Ubuntu 14.04 and newer and also on Mac OS X.
The intent is to also re-use these Travis CI scripts to build Scopy locally on Linux.

Some points:
* boost is no longer required to be built ; Ubuntu 14.04 has boost 1.55 which is sufficient for GnuRadio
* for Ubuntu 18.04 the list of packages required to be built has been reduced
* most patch files have been moved inside the build scripts
* the build steps for packages have been wrapped in functions to make things a bit more compact


CI/travis/before_install_darwin.sh
```diff
@@ -1,171 +1,86 @@
...
-  tar -xzvf libsigrokdecode-0.4.1.tar.gz
-  cd libsigrokdecode-0.4.1
-  ./configure
-  make >/dev/null
+
+QMAKE="$(command -v qmake)"
+
+pip install --user cheetah
+
+make_build_wget "libsigrok-0.5.0" "http://sigrok.org/download/source/libsigrok/libsigrok-0.5.0.tar.gz" "" "" "./autogen.sh"
```
AlexandraTrifan: Could we change this to get the following libsigrok version: https://github.com/sigrokproject/libsigrok

commodo: ack ; will try ;
i forgot about it :)

thanks

commodo:
if no objections, will merge this tomorrow

PR #476, opened 2018-08-03
build: fix Linux build (part 2)

commodo:
The build on master seems to fail currently, because of some fluke/goof with how caching works on Travis CI.

Link: 
https://travis-ci.org/analogdevicesinc/scopy/builds/411606380

Fixed this by patching dependencies only once when they're downloaded.
This also simplifies things on patching in general.

Also fixes the local build of `libad9361-iio` when building Scopy.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ build
 deps
 scopy
 *.swp
+html/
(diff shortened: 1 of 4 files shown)


commodo:
all green now;
let's see how it goes on master

PR #479, opened 2018-08-06
build,travis-ci: add Xenial builds to the job run

commodo:
Time to also build for Ubuntu 16.04.
Unfortunately, the QT5 libs are still too old for Scopy, so we need to pull
an external Qt5.9 repo from launchpad (similar as for Trusty).

Link is:
  https://launchpad.net/~beineri/+archive/ubuntu/opt-qt592-xenial

Same packager also has builds for Xenial.

And we also need to specifically pull the Mesa GL dev libs so that libqwt
builds fine.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,6 +25,10 @@ matrix:
       os: linux
       dist: trusty
       env: LDIST=-trusty
+    - compiler: gcc
+      os: linux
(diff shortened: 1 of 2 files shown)


commodo:
Will merge later today if no objections

PR #487, opened 2018-08-26
Compiler warnings

dNechita:
- Enable the -Wall set of warnings. This increase the awareness of possible coding issues.
 - Tell the compiler to treat a couple of these warnings as errors to ensure the warnings don't get overlooked.

Ideally all warnings should be treated as errors or at least a large part of them.  I've added only a a few as there were many unused functions/variables warnings as well as reorder warnings which would require too many code changes for a single PR.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -271,6 +271,14 @@ configure_file(Info.plist.cmakein ${CMAKE_CURRENT_BINARY_DIR}/Info.plist @ONLY)
 # Compiler options
 target_compile_options(${PROJECT_NAME} PUBLIC -Wall)
 
+#List of warnings to be treated as errors
+target_compile_options(${PROJECT_NAME} PUBLIC
+	-Werror=return-type
+	-Werror=uninitialized
+	-Werror=init-self
+	-Werror=switch
+)
+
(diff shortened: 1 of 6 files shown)


CMakeLists.txt
```diff
@@ -271,6 +271,14 @@ configure_file(Info.plist.cmakein ${CMAKE_CURRENT_BINARY_DIR}/Info.plist @ONLY)
 # Compiler options
 target_compile_options(${PROJECT_NAME} PUBLIC -Wall)
 
+#List of warnings to be treated as errors
+target_compile_options(${PROJECT_NAME} PUBLIC
+        -Werror=return-type
```
commodo: nitpick: indentation inconsistency here

src/pg_patterns.cpp
```diff
@@ -1300,6 +1300,9 @@ uint16_t UARTPattern::encapsulateUartFrame(char chr, uint16_t *bits_per_frame)
 		parity_bit_available = true;
 		parity_bit_value = 0;
 		break;
+
+	case SP_PARITY_INVALID:
+		qDebug() << "Invalid parity setting detected";
```
commodo: this patch looks like it could be in it's own commit

commodo:
rest looks good

PR #537, opened 2019-02-27
Breakpad implementation

adisuciu:
This pull request handles two things:

1. Implements breakpad crash handler. Whenever Scopy crashes, the breakpad crash exception handler should capture the exception and create a dump file which can be analyzed. Appveyor scripts also provide debug symbols for analysis.

2.  Implementing this caused appveyor 1hour build time to be exceded. I split the job into two different environments. One for 32bit and one for 64bit, so build time is reduced by half.


src/osc_adc.cpp
```diff
@@ -376,7 +376,6 @@ void M2kAdc::setSampleRate(double sr)
 	if(filteringEnabled())
 	{
 		GenericAdc::setSampleRate(sr);
-		setOversamplingRatio(1);
```
DanielGuramulta: This line is necessary to prevent the bug fixed in: https://github.com/analogdevicesinc/scopy/commit/2fc96ccec862f21d0706990a31a028e3ca544c42

src/oscilloscope.cpp
```diff
@@ -4113,8 +4113,6 @@ void Oscilloscope::writeAllSettingsToHardware()
 
 		/*iio_device_attr_write_longlong(adc->iio_adc_dev(),
 			"oversampling_ratio", 1);*/
-
-		m2k_adc->setSampleRate(active_sample_rate);
```
DanielGuramulta: This line fixed a bug where the sample rate would not be set when disabling/enabling the filter and playing with the spectrum analyzer. Fix introduced in commit: https://github.com/analogdevicesinc/scopy/commit/2fc96ccec862f21d0706990a31a028e3ca544c42

src/main.cpp
```diff
@@ -96,5 +124,9 @@ int main(int argc, char **argv)
 				 Q_ARG(QString, script));
 	}
 
-	return app.exec();
+
+	int ret = 0;
+	ret = app.exec();
+	return ret;
```
DanielGuramulta: I see no reason to create a temporary ret int. This can be left as "return app.exec()"

CMakeLists.txt
```diff
@@ -256,7 +268,7 @@ target_link_libraries(${PROJECT_NAME} LINK_PRIVATE
 		${LIBSIGROK_DECODE_LIBRARIES}
 		${GLIBMM_LIBRARIES}
 		${SIGCPP_LIBRARIES}
-		${GLIB_LIBRARIES}
+		${GLIB_LIBRARIES}
```
DanielGuramulta: Unnecessary tabs after "${GLIB_LIBRARIES}"

src/scopyApplication.cpp
```diff
@@ -0,0 +1,99 @@
...
+		QFile::rename(dumpFullPath, prevCrashDump);
+
+
+	}
+
+	QByteArray ba = qd.path().toLocal8Bit();
+	const char *appDirStr = ba.data();
+
+#ifdef Q_OS_LINUX
+	descriptor = new google_breakpad::MinidumpDescriptor(qd.path().toStdString().c_str());
```
dNechita: You could use unique pointer here. This way you can avoid the #ifdef q_os_linux in destructor and even skip defining the desctructor.

src/scopyApplication.hpp
```diff
@@ -0,0 +1,50 @@
...
+#define __STDC_FORMAT_MACROS
+#ifdef Q_OS_WIN
+#include "breakpad/client/windows/handler/exception_handler.h"
+#endif
+
+#ifdef Q_OS_LINUX
+#include "client/linux/handler/exception_handler.h"
+#endif
+
+//#ifdef CATCH_UNHANDLED_EXCEPTIONS
```
dNechita: needed?

src/osc_adc.cpp
adisuciu: Dropped commit that introduced this

appveyor.yml
```diff
@@ -1,96 +1,147 @@
...
     # Install an older version of icu library
-    - C:\msys64\usr\bin\bash -lc "pacman -U --noconfirm http://repo.msys2.org/mingw/i686/mingw-w64-i686-icu-58.2-3-any.pkg.tar.xz"
-    - C:\msys64\usr\bin\bash -lc "pacman -U --noconfirm http://repo.msys2.org/mingw/i686/mingw-w64-i686-icu-debug-libs-58.2-3-any.pkg.tar.xz"
-    - C:\msys64\usr\bin\bash -lc "pacman -U --noconfirm http://repo.msys2.org/mingw/x86_64/mingw-w64-x86_64-icu-58.2-3-any.pkg.tar.xz"
-    - C:\msys64\usr\bin\bash -lc "pacman -U --noconfirm http://repo.msys2.org/mingw/x86_64/mingw-w64-x86_64-icu-debug-libs-58.2-3-any.pkg.tar.xz"
+    - C:\msys64\usr\bin\bash -lc "pacman -U --noconfirm http://repo.msys2.org/mingw/%ARCH%/mingw-w64-%ARCH%-icu-58.2-3-any.pkg.tar.xz"
+    - C:\msys64\usr\bin\bash -lc "pacman -U --noconfirm http://repo.msys2.org/mingw/%ARCH%/mingw-w64-%ARCH%-icu-debug-libs-58.2-3-any.pkg.tar.xz"
+
+    # Install olderversion of binutils
+    #- C:\msys64\usr\bin\bash -lc "pacman -U --noconfirm http://repo.msys2.org/msys/%ARCH%/binutils-2.28-1-%ARCH%.pkg.tar.xz"
```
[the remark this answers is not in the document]
adisuciu: Cleaned up appveyor.yml with commit 2d88f32

src/scopyApplication.cpp
adisuciu: I'd rather not use complex objects (such as unique_ptr) when dealing with these kinds of exceptions as they operate very low level.

src/scopyApplication.hpp
adisuciu: Enabling this will catch unhandled exceptions but it seems OS-dependant. I'd keep this as it will make debug easier in the future.

src/scopyApplication.cpp
dNechita: ok

src/scopyApplication.hpp
dNechita: Actually, what I wanted to say was that I've seen this in the .cpp file as well and was wondering if is is needed in the header as well.

adisuciu: This was supposed to be a  #define.

PR #544, opened 2019-03-11
Oscilloscope math channel fixes

AlexandraTrifan:
1) This pull request adds a fix for issue #539 . Adding a math channel that results from a division has undefined behaviour (can generate "nan" or infinity/-infinity values ). With this PR, a fix for this scenario is added. 
- On the measurement side, we skip nan values. 
- On the display/plot side, we add a paint filter attribute to handle nan values.
- Output values from math channels are limited to a known value (both positive and negative, which depends on the attenuation for each channel) .

2) On the same flow for the oscilloscope, another fix was added in this PR. Math channels use as inputs the 1st and 2nd channel. If the channels have an attenuation different from 1X, this should be taken into consideration when adding math channels. 
In order to do this, a multiply block is added as input for the math channels. The multiply block uses as constant value the value of the probe attenuation for each channel.

Signed-off-by: Alexandra.Trifan <Alexandra.Trifan@analog.com>

### DIFF
--- a/src/TimeDomainDisplayPlot.h
+++ b/src/TimeDomainDisplayPlot.h
@@ -111,6 +111,10 @@ public:
 	unsigned long long channelsDataLength, bool curvesAttached = true);
   bool unregisterSink(std::string sinkName);
 
+  bool registerMathWaveform(std::string sinkUniqueNme, unsigned int numChannels,
+	unsigned long long channelsDataLength, bool curvesAttached = true);
+  bool unregisterMathWaveform(std::string sinkName);
+
   long dataStartingPoint() const;
 
   void addZoomer(unsigned int zoomerIdx);
@@ -185,6 +189,7 @@ protected:
   QVector<QVector<double>> d_preview_ydata;
   QVector<QwtPlotCurve *> d_preview_curves;
   bool isReferenceWaveform(QwtPlotCurve *curve);
+  bool isMathWaveform(QwtPlotCurve *curve);
   int countReferenceWaveform(int position);
 
 private:
@@ -220,6 +225,7 @@ private:
   QColor getChannelColor();
 
   QMap<QString, QwtPlotCurve *> d_ref_curves;
+  QMap<QString, QwtPlotCurve *> d_math_curves;
   int d_nb_ref_curves;
   int getCurveNextTo(int pos);
 };
--- a/src/TimeDomainDisplayPlot.h
+++ b/src/TimeDomainDisplayPlot.h
@@ -111,9 +111,9 @@ public:
 	unsigned long long channelsDataLength, bool curvesAttached = true);
   bool unregisterSink(std::string sinkName);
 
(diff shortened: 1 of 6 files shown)


src/oscilloscope.cpp
```diff
@@ -3238,14 +3238,17 @@ void Oscilloscope::editMathChannelFunction(int id, const std::string& new_functi
...
-	triggerRightMenuToggle(
-		static_cast<CustomPushButton* >(ui->btnAddMath), false);
-	triggerRightMenuToggle(
-		static_cast<CustomPushButton* >(chn_widget->menuButton()), true);
-	math_pair->first.btnAddChannel->setText("Add Channel");
-	math_pair->second->setFunction("");
-	addChannel = true;
-	ch_ui->btnEditMath->setChecked(false);
+
+	if (new_function != chn_widget->function().toStdString()) {
```
dNechita: Wouldn't it make sens to check this at the beginning of the function? And in case the new function is the same as the current on, just return?

AlexandraTrifan: That's the way it was done. However, when the probe_attenuation changes (aprox line 679), we need to update the rail block based on the new values (which sets a max and a min for math output values). If this function is not called, the rail block will not update, and the results will not look good.  

I might be able to update the rail block on the run, without needing to disconnect and reconnect all the blocks. Let me look into it.

AlexandraTrifan: Added the check back in the function. We can update the rail max and min values at runtime, so there is no need to edit the math channel every time.

AlexandraTrifan:
Changelog V1 -> V2:

- Change the rail block max and min value at runtime.
- Register and unregister math waveforms separately, in order to be able to identify them,
- Add missing include for measurements.

src/TimeDomainDisplayPlot.h
```diff
@@ -111,6 +111,10 @@ class TimeDomainDisplayPlot: public DisplayPlot
 	unsigned long long channelsDataLength, bool curvesAttached = true);
   bool unregisterSink(std::string sinkName);
 
+  bool registerMathWaveform(std::string sinkUniqueNme, unsigned int numChannels,
```
dNechita: First param should be const&. Same for the unregister method.

src/TimeDomainDisplayPlot.h
```diff
@@ -185,6 +189,7 @@ private Q_SLOTS:
   QVector<QVector<double>> d_preview_ydata;
   QVector<QwtPlotCurve *> d_preview_curves;
   bool isReferenceWaveform(QwtPlotCurve *curve);
+  bool isMathWaveform(QwtPlotCurve *curve);
```
dNechita: Method can be const.

AlexandraTrifan:
Changelog V2 -> V3:

- Use a Qt method for nan, instead of isnan() (which is not really cross platform)
- Change some method declarations.

PR #545, opened 2019-03-11
Digital instruments: Disable digital decoders support.

AlexandraTrifan:
This PR implements an option to choose whether the Scopy instance loads the digital decoders. The functionality may be limited for the Logic Analyzer and the Pattern Generator if the decoders are not loaded. 
There are 2 ways to disable the decoders:
- Using a command line option: **--nodecoders**
- Using the option listed in the preference panel . (In this case, Scopy needs to be reset manually.  Or should we automatically reset Scopy when the option is checked? This can be discussed here.)

### DIFF
--- a/src/pattern_generator.cpp
+++ b/src/pattern_generator.cpp
@@ -58,6 +58,7 @@
 #include "filter.hpp"
 #include "pattern_generator.hpp"
 #include "dynamicWidget.hpp"
+#include "tool_launcher.hpp"
 #include "pattern_generator_api.hpp"
 
 // Generated UI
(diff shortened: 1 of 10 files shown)


src/pattern_generator.cpp
```diff
@@ -448,6 +450,19 @@ void PatternGenerator::enableCgSettings(bool en)
 	colour_button_low->setEnabled(en);
 }
 
+bool PatternGenerator::getUse_decoders() const
```
dNechita: I think it is a good idea not to mix two naming conventions (camel-case and snake-case). Also, the rest of the code is mostly written with camel-case. Keeping things consistent makes code easier to read.

src/preferences.cpp
```diff
@@ -496,3 +512,13 @@ void Preferences_API::setMiniHist(const bool &enabled)
 {
 	preferencePanel->mini_hist_enabled = enabled;
 }
+
+bool Preferences_API::getDigitalDecoders() const
+{
+	return preferencePanel->digital_decoders_enabled;
+}
+
+void Preferences_API::setDigitalDecoders(const bool &enabled)
```
dNechita: There is no benefit in passing a constant reference to a bool variable. It just makes the function signature a little bit longer.

AlexandraTrifan:
Changelog V1 -> V2:

- Implement required fixes.
- Add a popup to inform the user that the preference changes regarding the digital decoders will only apply after a reset.

PR #547, opened 2019-03-15
treewide: fix sign-compare warnings

commodo:
These changes fix sign-compare warnings reported by G++ 7.3.
This is not the complete lists. It is a set of changes identified up to this point in time, and more will be identified in the coming days.

### DIFF
--- a/src/apiObject.cpp
+++ b/src/apiObject.cpp
@@ -40,7 +40,7 @@ template <typename T> void ApiObject::save(QSettings& settings,
 {
 	settings.beginWriteArray(prop, list.size());
 
-	for (unsigned int i = 0; i < list.size(); i++) {
+	for (int i = 0; i < list.size(); i++) {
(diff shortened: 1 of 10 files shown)


src/adc_sample_conv.hpp
```diff
@@ -67,7 +67,7 @@ namespace adiscope {
 		void setHardwareGain(int connection, float gain);
 		float hardwareGain(int connection) const;
 
-		int work(int noutput_items,
+		int work(unsigned int noutput_items,
```
DanielGuramulta: This method here overrides work in [sync_block.h](https://github.com/gnuradio/gnuradio/blob/master/gnuradio-runtime/include/gnuradio/sync_block.h#L52) which is a pure virtual method, changing the signature of this method will cause errors.

commodo:
Changelog v1 -> v2:
* removed `work()` signature change

PR #550, opened 2019-03-15
src/plot_utils.hpp: define virtual destructor

commodo:
G++ complains in dbgraph.cpp:
```
scopy/src/dbgraph.cpp:200:9: error: deleting object of polymorphic class type ‘adiscope::PrefixFormatter’ which has non-virtual destructor might cause undefined behavior [-Werror=delete-non-virtual-dtor]
  delete formatter;

```


commodo:
CHangelog v1 -> v2:
* added explicit destrutor to PrefixFormatter class

Issue #529, opened 2019-02-04
Plot curve visualization options


adisuciu:
QwtPlot supports multiple types of data rendering. Currently we only render data using lines (or interpolation method). Implement other visualization methods (like point or sample/hold)

damercer:
I think in Scopy with the high sample rate of M2K and the limit set on the smallest time/div setting allowed, the sample points are closely spaced so that drawing straight lines between points is often good enough to render nice traces. However another rendering option you might look at is some sort of smooth curves like splines to connect the dots. Again, this only helps visually if the sample points are many pixels apart ( like having only a few samples per horizontal time division ).
Anyway, just a thought.
Doug

adisuciu:
Fixed with https://github.com/analogdevicesinc/scopy/pull/556

amiclaus:
On Scopy version: e28686c (v1.0.5) the feature is implemented

PR #556, opened 2019-03-26
Osc plot modes

DanielGuramulta:
This pr implements feature: #529

### DIFF
--- a/src/TimeDomainDisplayPlot.cc
+++ b/src/TimeDomainDisplayPlot.cc
@@ -681,6 +681,54 @@ TimeDomainDisplayPlot::stemPlot(bool en)
   }
 }
 
+void TimeDomainDisplayPlot::setPlotLineStyle(unsigned int chIdx, unsigned int style)
+{
+	if (chIdx < 0 || chIdx >= d_plot_curve.size()) {
+		return;
+	}
+
+	auto curve = d_plot_curve[chIdx];
+
+	switch (style) {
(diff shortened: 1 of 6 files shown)


src/TimeDomainDisplayPlot.h
```diff
@@ -93,6 +93,9 @@ class TimeDomainDisplayPlot: public DisplayPlot
 
   void stemPlot(bool en);
 
+  void setPlotLineStyle(unsigned int chIdx, unsigned int style);
+  int getLineStyle(unsigned int chIdx);
```
dNechita: Shouldn't the name be consistent between a setter and a getter? (either setPlotLineStyle + getPlotLineStyle or setLineStyle + getLineStyle)

src/TimeDomainDisplayPlot.cc
```diff
@@ -681,6 +683,67 @@ TimeDomainDisplayPlot::stemPlot(bool en)
...
+		break;
+		case 2:
+		curve->setStyle(QwtPlotCurve::CurveStyle::Steps);
+		replot();
+		break;
+		case 3:
+		curve->setStyle(QwtPlotCurve::CurveStyle::Sticks);
+		replot();
+		break;
+		case 4:
```
dNechita: So I can set style 4 but then the getter only returns styles 0 through 3. It doesn't seem alright.

Issue #536, opened 2019-02-26
The signal generator does not start both channels at the same time


adisuciu:
The signal generator does not start both channels at the same time.

FW - 0.15
![screenshot from 2019-02-26 13-37-22](https://user-images.githubusercontent.com/8676912/53410269-f8fd3980-39cb-11e9-9ffe-249f20dd14fe.png)

Fw - 0.17
![screenshot from 2019-02-26 13-29-19](https://user-images.githubusercontent.com/8676912/53410260-f3075880-39cb-11e9-8479-39fd9e5315f3.png)

dbogdan:
The issue will be solved by implementing gpio_chip->set_multiple() for https://github.com/analogdevicesinc/linux/blob/m2k/drivers/input/keyboard/adp5589-keys.c.

adisuciu:
Should be fixed with https://github.com/analogdevicesinc/scopy/pull/557

amiclaus:
On Scopy version: e28686c (v1.0.5), FW: v0.23 the bug is fixed.

PR #557, opened 2019-03-28
src/signal_generator: powerdown handling change

btogorean:
Related to issue #536

Enable output channels before start of DMA data output. In this way there is no delay between output channels.
Also add check and if channel is disabled from GUI don't release corresponding "powerdown" signal.

Tested on Windows 10 x64.

### DIFF
--- a/src/signal_generator.cpp
+++ b/src/signal_generator.cpp
@@ -1249,6 +1249,12 @@ void SignalGenerator::start()
 {
 	QVector<struct iio_channel *> enabled_channels;
 
+	if (amp1 && amp2) {
+		/* FIXME: TODO: Move this into a HW class / lib M2k */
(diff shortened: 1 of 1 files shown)


src/signal_generator.cpp
```diff
@@ -1412,19 +1412,18 @@ void SignalGenerator::stop()
 
 void SignalGenerator::startStop(bool pressed)
 {
+	if (amp1 && amp2) {
```
AlexandraTrifan: When a channel is enabled or disabled from the GUI, while the Signal Generator is running, this function does not get called again. 
If a channel is enabled/disabled while running, the method **stop**() is called, then **start**() to restart the acquisition. In this case, should we move the "powerdown" controls in the start() and stop() function?

PR #569, opened 2019-04-12
treewide: fix compiler warninges (series 2)

commodo:
Some bits have been commented out only [vs being removed].
That's because, there is code there that was commented-out instead of being removed.

### DIFF
--- a/src/adc_sample_conv.cpp
+++ b/src/adc_sample_conv.cpp
@@ -71,14 +71,14 @@ int adc_sample_conv::work(int noutput_items,
 		float *out = static_cast<float *>(output_items[i]);
 
 		if (inverse)
-			for (unsigned int j = 0; j < noutput_items; j++)
+			for (int j = 0; j < noutput_items; j++)
 				out[j] = convVoltsToSample(in[j],
 						d_correction_gains[i],
 						d_filter_compensations[i],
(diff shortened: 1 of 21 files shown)


src/signal_generator_api.cpp
```diff
@@ -23,7 +23,7 @@ QList<int> SignalGenerator_API::getMode() const
 {
 	QList<int> list;
 
-	for (unsigned int i = 0; i < gen->channels.size(); i++) {
+	for (int i = 0; i < gen->channels.size(); i++) {
```
dNechita: Replacing all these IFs with c++ range-based for loops will make things easier. There will be no need to ask yourself what type "i" variable should have since it will be replaced with an iterator that usually is declared as auto. Also if container changes, let's say from qt containers to std containers, range-based for will not have to change, while in the actual form, "i" will need to be declared unsigned because that is what size() of std container returns.

commodo:
Changelog v1 -> v2:
* removed commit `signal_generator_api: fix sign-compare warnings with index iterator`; that one may deserve a bit deeper thought

src/signal_generator_api.cpp
commodo: removed this commit for now;
will give this a deeper thought;

thanks
alex

PR #570, opened 2019-04-15
Debugger: avoid reading corrupt data from IIO attributes

AlexandraTrifan:
Increase the number of characters being read for an IIO attribute.

Buffer overflows might happen if we use the older value.

### DIFF
--- a/src/debug.cpp
+++ b/src/debug.cpp
@@ -19,6 +19,8 @@
 
 #include "debug.h"
 
+static const size_t maxAttrSize = 512;
+
 using namespace adiscope;
 
 Debug::Debug(QObject *parent) : QObject(parent)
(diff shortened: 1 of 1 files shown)


src/debug.cpp
```diff
@@ -19,6 +19,8 @@
 
 #include "debug.h"
 
+#define MAX_ATTR_SIZE 512
```
dNechita: It is a [good practice](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#es31-dont-use-macros-for-constants-or-functions) to not use macros for constants.
In this case `static const size_t maxAttrSize = 512;` would be more appropriate.

AlexandraTrifan: Thank you. 
Updated the PR.

PR #621, opened 2019-06-24
Calibrate button

adisuciu:


src/calibration.cpp
```diff
@@ -760,74 +764,65 @@ void Calibration::dacOutputDC(struct iio_device *dac,
...
 void Calibration::dacOutputStop()
 {
-	if (m_dac_a_buffer) {
-		iio_buffer_cancel(m_dac_a_buffer);
-		iio_buffer_destroy(m_dac_a_buffer);
-		m_dac_a_buffer = NULL;
-	}
 
-	if (m_dac_b_buffer) {
-		iio_buffer_cancel(m_dac_b_buffer);
```
dNechita: iio_buffer_cancel in no longer called in turnDACOutputOff(). It this ok?

adisuciu: You are correct, it should be called. I'll add this.

Issue #640, opened 2019-07-10
Scopy exits when generating log(t) 


andreeaipop:
Signal generator: Math : log(t)
For frequencies under 500kHz  of the log(t) waveform Scopy app exits.

adisuciu:
Fixed with https://github.com/analogdevicesinc/scopy/pull/646

andreeaipop:
On Scopy build: d198b9a (v1.0.6) the bug is fixed.

PR #646, opened 2019-07-12
SignalGenerator: Prevent crash when plotting infinite values

adisuciu:
This situation can occur when using math function generator
Fixes https://github.com/analogdevicesinc/scopy/issues/640

### DIFF
--- a/src/signal_generator.cpp
+++ b/src/signal_generator.cpp
@@ -71,6 +71,9 @@
 
 #define AMPLITUDE_VOLTS	5.0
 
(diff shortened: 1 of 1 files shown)


src/signal_generator.cpp
```diff
@@ -71,6 +71,9 @@
 
 #define AMPLITUDE_VOLTS	5.0
 
+#define MAX_PREVIEW_RANGE (float)(SHRT_MAX)
```
DanielGuramulta: Avoid using macros for constants, you could use instead:
`static const float MAX_PREVIEW_RANGE`

Issue #633, opened 2019-07-04
Cannot import different waveforms on SG channels


amiclaus:
Scopy build: 1f47a67

Load CSV file into Signal Generator on channel 1. 

i.e. 
[stairstep.zip](https://github.com/analogdevicesinc/scopy/files/3359688/stairstep.zip)

Now import another CSV file on channel 2 with different waveform type.

[square.zip](https://github.com/analogdevicesinc/scopy/files/3359709/square.zip)

The waveform on channel 1 changes to the latest imported waveform.

PR #647, opened 2019-07-12
SignalGenerator: Fix loading two buffers in the signal generator

adisuciu:
Fixes https://github.com/analogdevicesinc/scopy/issues/633


src/pg_patterns.cpp
```diff
@@ -3258,6 +3258,22 @@ Pattern *ImportPatternUI::get_pattern()
...
+{
+	loadFileData();
+	parse_ui();
+}
+
+void ImportPatternUI::loadFileData()
+{
+
+	FileManager fm("Pattern Generator");
+	fm.open(pattern->fileName, FileManager::IMPORT);
```
DanielGuramulta: fm.open() might throw an exception if the file is corrupt, or it can't be opened. There should be a try-catch block here to make sure Scopy won't crash when this happens.

adisuciu: Fixed

PR #660, opened 2019-07-25
Install older version of hdf5 and libmatio

DanielGuramulta:
Signed-off-by: Daniel Guramulta <Daniel.Guramulta@analog.com>


rgetz:
Shouldn’t we be updating things so it’s compatible with the latest version? Otherwise we will end up with loads of things that’s are never maintainable...

DanielGuramulta:
Managed to get the latest qt version working for appveyor. The latest version of boost is also working now for gnuradio (https://github.com/analogdevicesinc/gnuradio/commit/025ac604d9d06f9dbe902e7c1cd0c661b8eec55c). The only problem is that hdf5 is not up to date on the msys2 repo and the latest available there is not working with the latest matio. Until hdf5 is updated for msys2 we are stuck with this older version. When things get updated we will only revert https://github.com/analogdevicesinc/scopy/pull/660/commits/b5aec9f0e84e3ccd61f183042c509cf271020332. After this gets merged(https://github.com/analogdevicesinc/scopy-mingw-build-deps/pull/2), I will remove https://github.com/analogdevicesinc/scopy/pull/660/commits/e010dd0899b88ba35637a6d36bd19887595b729b and this pr is good to go.

PR #664, opened 2019-08-06
Spectrum reference waveform

DanielGuramulta:


adisuciu:
When snapshotting / exporting also save channel parameters such as averaging/window/etc and show them in the reference/snapshot widget.

PR #668, opened 2019-08-12
build,travis: make ubuntu version check more robust

commodo:
The version check functions were picked up from:
  http://ask.xmodulo.com/compare-two-version-numbers.html

The previous check would only check for Ubuntu Bionic (18.04) and disregard
18.10 & 19.04.

This check should work to include anything starting with 18.04 and newer.

### DIFF
--- a/CI/travis/lib.sh
+++ b/CI/travis/lib.sh
@@ -12,13 +12,26 @@ fi
 
 export PYTHON3=python3
 
+version_gt() { test "$(echo "$@" | tr " " "\n" | sort -V | head -n 1)" != "$1"; }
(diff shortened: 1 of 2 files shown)


CI/travis/lib.sh
```diff
@@ -12,13 +12,26 @@ fi
...
+
+get_dist_id() {
+	lsb_release -i -s
+}
+
+get_version() {
+	lsb_release -r -s
+}
+
+version_gt() {
```
dNechita: Isn't this a redefinition of the version_gt() above?

commodo: yep; thanks for catching;
i think i tried too many things at once and got that from 2 different sources and forgot to remove one

commodo:
Changelog v1 -> v2:
* removed duplicate  `version_gt` function

PR #682, opened 2019-08-21
SpectrumAnalyzer: hide averaging value when averaging type is Sample

DanielGuramulta:
Signed-off-by: Daniel Guramulta <daniel.guramulta@analog.com>

### DIFF
--- a/src/spectrum_analyzer.cpp
+++ b/src/spectrum_analyzer.cpp
@@ -1082,6 +1082,10 @@ void SpectrumAnalyzer::on_comboBox_type_currentIndexChanged(const QString& s)
 		return;
 	}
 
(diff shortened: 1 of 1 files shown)


src/spectrum_analyzer.cpp
```diff
@@ -1082,6 +1082,15 @@ void SpectrumAnalyzer::on_comboBox_type_currentIndexChanged(const QString& s)
 		return;
 	}
 
+	if ((*it).second == FftDisplayPlot::AverageType::SAMPLE) {
```
dNechita: You could write this in a more compact way: setVisible(currentIndex==SAMPLE)

PR #717, opened 2019-09-27
Logic Analyzer: Fix data export

AlexandraTrifan:
When Export is pressed the Logic Analyzer is paused, but if nothing is
selected (or Cancel is pressed), it does not resume the capture. This
brings the tool in an unknown state.

### DIFF
--- a/src/logic_analyzer.cpp
+++ b/src/logic_analyzer.cpp
@@ -685,7 +685,6 @@ QString LogicAnalyzer::saveToFile()
 	QString endRow = "\n";
 	QString selectedFilter;
 	bool done = false;
-	bool paused = false;
 	bool noChannelEnabled = true;
 	
 	exportConfig = exportSettings->getExportConfig();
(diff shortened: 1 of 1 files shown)


src/logic_analyzer.cpp
```diff
@@ -710,8 +710,11 @@ QString LogicAnalyzer::saveToFile()
 	    &selectedFilter, (m_useNativeDialogs ? QFileDialog::Options() : QFileDialog::DontUseNativeDialog));
 
 
-	if(fileName.isEmpty())
+	if(fileName.isEmpty()) {
```
commodo: curios: is there a preferred style here with respect to `if()` vs `if ()`
the file does not seem consistent, so i don't know which should be used;

src/logic_analyzer.cpp
```diff
@@ -732,6 +735,8 @@ QString LogicAnalyzer::saveToFile()
 	} else {
 		QFile file(fileName);
 		if( !file.open(QIODevice::WriteOnly)) {
+			if(paused)
+				startStop(true);
```
commodo: as a "hot-fix" this looks fine, but as a general mechanism, i would see about maybe moving the `startStop()` outside of this function;
saving-to-file should not typically do starting/stopping of a state-machine

AlexandraTrifan: You are right, i will restructure it.

src/logic_analyzer.cpp
```diff
@@ -894,10 +886,21 @@ bool LogicAnalyzer::exportTabCsv(QString separator, QString filename)
...
+	if(m_running) {
+		paused = true;
+		startStop(false);
+	}
+
 	QString filename = saveToFile();
+
+	if(paused) {
+		startStop(true);
+	}
```
commodo: curios: what is the style-guide for single-block if statements?
depending on who you ask, some projects recommend not adding curly braces for them
so:
```
if(paused)
    startStop(true);
```

AlexandraTrifan: We are using curly braces for single-block if statements. But the old code (before the style-guide was put in place) was not cleaned up / updated yet.

commodo: ack
thanks for the info :)

btw: how much is the fish?

PR #739, opened 2019-12-11
Logic Analyzer: fix blocking when running

DanielGuramulta:
When running the LogicAnalyzer with multiple groups enabled, each having a
decoder active, Scopy would freeze and never recover. The problem is that
each instance of DecoderStack (the class that manages decoding) spawns a
thread that would lock a global mutex for the access to libsigrokdecode.
When this decoders are stopped if the thread that currently owns access to
libsigrokdecode is not joined first, the freeze occurs as there is no way
to interrupt that thread from waiting to acquire the mutex, while also not
having a way to interrupt the thread that acquired the mutex. To fix this
we allow each thread to defer the mutex while only one of them gets to own
it at a time. When the threads are now joined, they will stop trying to
acquire the mutex and exit peacefully.

Signed-off-by: Daniel Guramulta <daniel.guramulta@analog.com>

### DIFF
--- a/src/pulseview/pv/data/decoderstack.cpp
+++ b/src/pulseview/pv/data/decoderstack.cpp
@@ -388,7 +388,14 @@ void DecoderStack::decode_proc()
 	assert(segment_);
 
 	// Prevent any other decode threads from accessing libsigrokdecode
-	lock_guard<mutex> srd_lock(global_srd_mutex_);
+	unique_lock<mutex> srd_lock(global_srd_mutex_, std::defer_lock);
+	do {
+		std::this_thread::sleep_for(std::chrono::milliseconds(1));
+	} while (!interrupt_ && !srd_lock.try_lock());
+
+	if (interrupt_) {
+		return;
+	}
 
 	// Create the session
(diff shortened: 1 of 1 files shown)


src/pulseview/pv/data/decoderstack.cpp
```diff
@@ -387,8 +387,15 @@ void DecoderStack::decode_proc()
 
 	assert(segment_);
 
-	// Prevent any other decode threads from accessing libsigrokdecode
-	lock_guard<mutex> srd_lock(global_srd_mutex_);
+    // Prevent any other decode threads from accessing libsigrokdecode
```
commodo: whitespace is inconsistent with the rest of the file

commodo:
rest looks good

src/pulseview/pv/data/decoderstack.cpp
DanielGuramulta: Indeed. I played with the settings of my IDE for another project, and forgot about that. Everything should be fine now. Thanks for noticing!

PR #746, opened 2019-12-19
Add digital filters to the ADC data acquisition flow

adisuciu:
Filters are only enabled for lowgain mode, as high gain should be calibrated using the trimmers on the M2k. There is an option to show/hide the controls in the oscilloscope channels. The values for the filters are saved in the Scopy.ini file.

### DIFF
--- a/src/frequency_compensation_filter.h
+++ b/src/frequency_compensation_filter.h
@@ -38,14 +38,14 @@ namespace adiscope {
        * \param itemsize the item size of the stream
        * \param nstreams number of streams to combine into a vector (vector size)
        */
-      static sptr make(bool enable = true, float TC = 1, float gain = -1,
+      static sptr make(bool enable = true, float TC = 1, float gain = 0,
 		       float sample_rate = 100000000);
-      virtual void set_enable(bool en) = 0;
-      virtual bool get_enable() = 0;
-      virtual void set_TC(float TC) = 0;
-      virtual float get_TC() = 0;
-      virtual void set_gain(float gain) = 0;
-      virtual float get_gain() = 0;
+      virtual void set_enable(bool en, int gain_mode = 2) = 0;
+      virtual bool get_enable(int gain_mode = 2) = 0;
+      virtual void set_TC(float TC, int gain_mode = 2) = 0;
+      virtual float get_TC(int gain_mode = 2) = 0;
+      virtual void set_filter_gain(float gain, int gain_mode = 2) = 0;
+      virtual float get_filter_gain(int gain_mode = 2) = 0;
       virtual void set_sample_rate(float sample_rate) = 0;
       virtual bool get_high_gain() = 0;
       virtual void set_high_gain(bool en) = 0;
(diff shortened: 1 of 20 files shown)


adisuciu:
Filters are now available for both low and high gain mode with different configurations

src/frequency_compensation_filter_impl.cc
```diff
@@ -0,0 +1,138 @@
...
+			out[i]=in[i]+(short)((out_f[i])*config[high_gain].gain);
+		}
+
+	}
+	else
+	{
+		memcpy(out,in,noutput_items*sizeof(short));
+	}
+	delete []out_f;
+        return noutput_items;
```
AlexandraTrifan: Indentation looks a bit different when comparing to other source files. Can we astyle the newly added files?

src/frequency_compensation_filter.h
```diff
@@ -0,0 +1,55 @@
...
+namespace adiscope {
+
+    class frequency_compensation_filter : virtual public gr::sync_block
+    {
+    public:
+      // gr::blocks::streams_to_vector::sptr
+      typedef boost::shared_ptr<frequency_compensation_filter> sptr;
+
+      /*!
+       * Make a stream-to-vector block.
```
AlexandraTrifan: Is this comment related to the GR block?

src/oscilloscope_api.cpp
```diff
@@ -639,6 +639,17 @@ QVariantList Oscilloscope_API::getChannels()
 	return list;
 }
 
+QVariantList Channel_API::getDigFilters() const
+{
+	QVariantList list;
+
+	for (Channel_Digital_Filter_API *each : digFilters)
```
AlexandraTrifan: missing "{" and  "}" ?

src/oscilloscope_api.hpp
```diff
@@ -308,10 +343,21 @@ class Oscilloscope_API : public ApiObject
...
+
 public:
 	explicit Channel_API(Oscilloscope *osc) :
-		ApiObject(), osc(osc) {}
-	~Channel_API() {}
+		ApiObject(), osc(osc) {
+			for(auto i=0;i<2;i++)
+				digFilters.append(new Channel_Digital_Filter_API(osc,this,i));
+		}
+
```
AlexandraTrifan: missing "{" and "}" ?

src/frequency_compensation_filter_impl.cc
adisuciu: fixed

src/frequency_compensation_filter.h
adisuciu: fixed

src/oscilloscope_api.cpp
adisuciu: fixed

src/oscilloscope_api.hpp
adisuciu: fixed

PR #761, opened 2020-01-22
Fix scopy freeze 4k

DanielGuramulta:


commodo:
curios: does this need to go into master?
or is it there already?

DanielGuramulta:
> 
> 
> curios: does this need to go into master?
> or is it there already?

The master branch is frozen as we are currently in the makings of a new release (v1.1.2). All the bug fixes are going into v1.1.2 and at the end when the release will be made, v1.1.2 will go into master.

PR #762, opened 2020-01-22
General: fix compiling of translations on Windows

DanielGuramulta:
Scopy would not compile on Windows due to the Linux style paths that are hardcoded in the command generating the .qm files
Using qt5_create_translation the problem no longer occurs as we don't invoke any commands like: mkdir, mv with Linux paths

Signed-off-by: Daniel Guramulta <Daniel.Guramulta@analog.com>


adisuciu:
These modifications were included with  https://github.com/analogdevicesinc/scopy/pull/771

PR #782, opened 2020-04-22
SignalGenerator: fix trapezoidal wave frequency rounding error

cristina-suteu:
fixes frequency  approximation bug when switching to trapezoidal waveform

the computed frequency is not displayed correctly causing waveforms to appear
out of sync.
Rounding the frequency using 3 decimals fixes this issue
as 3 decimals are also displayed in the spinbox.

Signed-off-by: Cristina Suteu <cristina.suteu@analog.com>

### DIFF
--- a/src/signal_generator.cpp
+++ b/src/signal_generator.cpp
@@ -836,7 +836,8 @@ void SignalGenerator::dutyChanged(double value)
 void SignalGenerator::trapezoidalComputeFrequency()
 {
 	auto ptr = getCurrentData();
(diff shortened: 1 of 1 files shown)


src/signal_generator.cpp
```diff
@@ -33,6 +33,7 @@
 #include <QPalette>
 #include <QSharedPointer>
 #include <QElapsedTimer>
+#include <QDebug>
```
DanielGuramulta, line 36: is this include still needed?

cristina-suteu: fixed

PR #787, opened 2020-05-07
Preferences: add translation feature

cristina-suteu:
Scopy translation feature
auto option for setting app to os language, if available
browse option for custom translation files
resources/languages directory for bundled translations


src/main.cpp
```diff
@@ -111,19 +112,34 @@ int main(int argc, char **argv)
...
+    const QStringList content = in.readAll().split("\n");
 	QString language="default";
-	for(auto s : content)
-	{
-		if(s.startsWith("language"))
-		{
-			language=s.split("=")[1];
+	for(auto s : content) {
+		if(s.startsWith("language")) {
+			language=s.split("=")[1].trimmed();
```
DanielGuramulta, line 119: From what I understand in this lines, the ini file, `Preferences.ini`, is manually opened and parsed. I think the language should be retrieved through the Preferences API.

src/main.cpp
```diff
@@ -111,19 +112,34 @@ int main(int argc, char **argv)
...
+	else if(languages.contains(language+".qm")) {
+		languageFileName=QDir(QCoreApplication::applicationDirPath()+"/resources/languages/"+language+".qm").path();
+	}
+	else { 
+		languageFileName = language;
+	}
 
-	QString languageFileName=QDir(QCoreApplication::applicationDirPath()+"/resources/languages/"+language+".qm").path();
+	f.close();
 	myappTranslator.load(languageFileName);
```
DanielGuramulta, line 143: Wouldn't it be easier to just
```cpp
        if (translator->load(locale, QCoreApplication::applicationName(), "_", dir)) {
            app.installTranslator(translator);
        }
```
where dir would be the directory of the languages. In this way the line above will try to load `scopy_en` or `scopy_de`, any scopy_<language> depending on the language given by the locale. If no language is found in this directory then no translator should be installed.
The languages (*.qm files) should be compiled in a resource and attached to the executable. The current implementation won't work for flatpak and the paths to the translation files would need to be given manually.

src/main.cpp
```diff
@@ -173,3 +189,4 @@ int main(int argc, char **argv)
 	}
 	return app.exec();
 }
+
```
DanielGuramulta, line 192: nitpik: extra line

src/preferences.cpp
```diff
@@ -152,24 +156,68 @@ Preferences::Preferences(QWidget *parent) :
...
 	pref_api->load(settings);
-
 	connect(ui->languageCombo, &QComboBox::currentTextChanged, [=](QString lang) {
-		language = lang;
-		if (m_initialized)
-			ui->label_restart->setVisible(true);
-		else
-			m_initialized = true;
-		Q_EMIT notify();
+			if(lang=="browse"){
```
DanielGuramulta, line 167: Should format this code:
```cpp
if () {
    ...
} else {
    ...
}
```

src/preferences.cpp
```diff
@@ -152,24 +156,68 @@ Preferences::Preferences(QWidget *parent) :
...
+				if (m_initialized)
+					ui->label_restart->setVisible(true);
+				else
+					m_initialized = true;
+        			}
+			Q_EMIT notify();
 	});
 }
 
+QStringList Preferences::getOptionsList()
```
DanielGuramulta, line 198: This also needs formatting

src/preferences.cpp
```diff
@@ -198,6 +246,7 @@ void Preferences::notifyChange()
 
 void Preferences::showEvent(QShowEvent *event)
 {
+
```
DanielGuramulta, line 249: extra line

src/preferences.cpp
```diff
@@ -152,24 +156,68 @@ Preferences::Preferences(QWidget *parent) :
...
+QStringList Preferences::getOptionsList()
+{   
+	QFile f(getPreferenceIniFile());
+	f.open(QFile::ReadOnly);
+    QTextStream in (&f);
+	const QStringList content = in.readAll().split("\n");
+	QString language;
+	for(auto s : content) {
+		if(s.startsWith("language")){
+			language=s.split("=")[1].trimmed();
```
DanielGuramulta, line 207: I think there already is a property called language for this that uses `getLanguage`, `setLanguage`, is the parsing of the ini file needed?

adisuciu:
> This pull request seems to add a little more than just `add translation feature`. I see there are also some changes regarding CI, and Signal Generators `MATLAB` files behavior. I don't think this changes should be together.



> This pull request seems to add a little more than just `add translation feature`. I see there are also some changes regarding CI, and Signal Generators `MATLAB` files behavior. I don't think this changes should be together.

Those changes are already on master. PR just needs a rebase.

PR #828, opened 2020-07-22
CI: fix appveyor all builds

adisuciu:


CI/appveyor/install_ubuntu_deps.sh
```diff
@@ -13,13 +13,14 @@ QWTPOLAR_BRANCH=master # not used
 LIBSIGROK_BRANCH=master
 LIBSIGROKDECODE_BRANCH=master #not used
 
+QMAKE=/home/adi/Qt/5.15.0/gcc_64/bin/qmake
```
DanielGuramulta, line 16: this path seems hardcoded for your system

adisuciu: In the build guide it is mentioned that this needs to be changed. We won't run this on appveyor since we can't install a custom version of Qt on Ubuntu. 

Any suggestions on what this should be ?

DanielGuramulta: maybe a default value like $HOME/Qt/$QTVERSION where $HOME is a default env variable for ubuntu (hope I'm not wrong) and QTVERSION a default value. Both changeable through options given to the script. I think this way of providing options to the script when running it, instead of modifying the script is better.
example:
```
./install_ubuntu_deps.sh --qtversion 5.15.0
```
or even
```
./install_ubuntu_deps.sh --qtpath <path_to_qt_instalation>
```

An example of a script using options might be this: https://github.com/analogdevicesinc/aditof_sdk/blob/master/scripts/linux/setup.sh#L61

DanielGuramulta: Also another candidate for an option that can be specified could be the `WORKDIR`, from what I see in the script there is no possibility of changing where to install stuff, without `manually modifying` the script. I would also add this as an option. The general usage of the script should become something like: `./install_ubuntu_deps.sh [option1] [option2]` with documentation on this options(`--help` option) not on how to modify the script.

DanielGuramulta: Optionally the script could prompt the user with a message that explains what happens:
Will install "this" into "that" using "this version" of Qt. with a prompt (yes or no) to continue.
yes or exit prompt implemented here: https://github.com/analogdevicesinc/aditof_sdk/blob/master/scripts/linux/setup.sh#L46

adisuciu:
I added a single paramter to the build which is the Qt version. If there is a need to extend this script, we can do so at a later time.

PR #831, opened 2020-07-26
PatternGenerator: Fixes I2C Read termination, I2C SCL/SDA timings and…

electronpilot:
… frequency

Summary of changes:
* I2C reads now terminate with a NACK.
* SCL edges changed so that they do not coincide with SDA edges (which can cause spurious start/stop conditions)
* The sampling rate has been tweaked to reflect the actual SCL frequency

Signed-off-by: Daniel Sadler <d.f.sadler@gmail.com>

The first item fixes an issue [my previous pull request](https://github.com/analogdevicesinc/scopy/pull/778) created - I had neglected to NACK the final byte of an I2C read.

The Appveyor config was a little broken at the time of submitting this. I managed to get it running using the changes in dd628fb0e7dbb2ef888fd5bf73bf5622571f9fe4 for testing (an MSYS2 contributor key was [causing issues](https://www.msys2.org/news/)), but I excluded those from this PR - I figured you would probably want to fix it properly yourselves. Feel free to use my fix if it helps.

### DIFF
--- a/src/patterngenerator/patterns/patterns.cpp
+++ b/src/patterngenerator/patterns/patterns.cpp
@@ -1689,13 +1689,13 @@ I2CPattern::I2CPattern()
 
 uint32_t I2CPattern::get_min_sampling_freq()
 {
-	return clkFrequency * (2);
+	return clkFrequency * (4);
 }
 
 uint32_t I2CPattern::get_required_nr_of_samples(uint32_t sample_rate,
 		uint32_t number_of_channels)
 {
-	auto samples_per_bit = 2*(sample_rate/clkFrequency);
+	auto samples_per_bit = 1*(sample_rate/clkFrequency);
 	auto IFS=interFrameSpace*samples_per_bit;
 
 	// size = samples/bit * (IFS+start(2), address(7), ack(1), (data(8) + ack(1))*data_len, stop(2)+IFS)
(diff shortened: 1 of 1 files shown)


adisuciu:
Thanks again ! 

I created a diff screenshot - top is master, bottom is this pull request
image

I put little circles near the differences - sampling points for SDA, NACK in the end to stop repeated read, total time generated for 5kHz I2C frequency.

If there are no comments, I'll merge later today

-Adrian

andreeaipop:
Hello,
I noticed in @adisuciu 's screenshoot that the last clock period in the pull-request version is different than the last clock period generated in the master version. See the picture below.
image
 Is this an issue, since the data is done reading?

electronpilot:
Good spot! 

In theory it shouldn't matter, the I2C spec allows clock stretching where the master or slave can drag out the Low clock as long as it likes even partway through a communication e.g. if the receiving device needs more time to process incoming data.

The I2C spec doesn't have a max duration for SCL LOW ([page 48-50](https://www.nxp.com/docs/en/user-guide/UM10204.pdf)); the only possible issue here might be around ensuring sufficient time to setup the STOP condition (the duration between SCL HIGH and SDA HIGH at the end). As far as I can tell this is the same as the minimum SCL High duration, which I have used here (2 samples).

Anything quicker than 100kHz Standard mode I2C technically needs more attention paying to timings (e.g. SCL low is roughly twice as long as SCL high for Fast mode), but since this wasn't implemented already I havent included any logic dependent on the chosen frequency. Wouldn't be hard to do at some point though.

adisuciu:
Thanks for looking it up. I merged it ! 
-Adrian

PR #838, opened 2020-07-30
La pg fixes

DanielGuramulta:
Some fixes for the Logic Analyzer and Pattern Generator


DanielGuramulta:
@adisuciu  fixed the conflict in the .ui file!

PR #850, opened 2020-08-19
Mixed signal view

DanielGuramulta:


DanielGuramulta:
@adisuciu @AlexandraTrifan could you give this a try and see how it works? Still has some issues but should be working with no problem. Make sure you have the latest libm2k and gr-m2k

adisuciu:
Needs rebase.

Still getting some crashes, and sometimes there is a desync in mixed signal view. If the fixes are not trivial, we should merge now and create PRs for bugfixes. @DanielGuramulta ?

adisuciu (approved):

PR #853, opened 2020-08-27
Oscilloscope: Disable mixed signal view.

teoperisanu:
Disable mixed signal view when using an incompatible firmware.


DanielGuramulta (approved):
Looks great! Will merge it later, still have some local changes to push on the mixed-signal-view branch

PR #855, opened 2020-08-28
Oscilloscope: change unit from Hz to sps for sample rate in osc plot

adisuciu:


rgetz:
Thanks - sample rate of a sampling system is "samples per second", not "Hertz". (Hz normally refers to a continuous analog frequency, or bandwidth, not a discrete rate).

PR #867, opened 2020-09-14
Appveyor MacOS: Update Qt to 5.15.1 for MacOS builds.

AlexandraTrifan:
We locked the builds to Qt 5.14.2 because Qt introduced a bug in Qt 5.15.0,
related to QPushButtons. Installing an older Brew formula using the github
Formula file was deprecated in the last couple of days and stopped working
for our build. Building the desired 5.14.2 version from source would take
too much time and would fail on our Appveyor builds.
Fortunately, Brew added today the 5.15.1 Formula on the official
repositories, so we can now use this updated version (this version
introduces the QPushButton fix that we needed.)


DanielGuramulta (approved):
👍

PR #871, opened 2020-09-18
Logic fixes

DanielGuramulta:
1. fixes regarding crashes when disabling the mixed signal view
2. when a new group is created select the first channel in it
3. fix the position on decoders in the mixed signal view menu
4. add information about how mixed signal view affects scopy when enabled
5. disable some trigger settings that won't work with the mixed signal view
6. make sure the stacked decoder button is correctly shown and set

### DIFF
--- a/src/trigger_settings.cpp
+++ b/src/trigger_settings.cpp
@@ -168,6 +168,7 @@ TriggerSettings::TriggerSettings(M2kAnalogIn* libm2k_adc,
 	MouseWheelWidgetGuard *wheelEventGuard = new MouseWheelWidgetGuard(this);
 	wheelEventGuard->installEventRecursively(this);
 
+	ui->mixedSignalLbl->setVisible(false);
 }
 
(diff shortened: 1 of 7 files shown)


src/trigger_settings.cpp
```diff
@@ -247,6 +248,20 @@ void TriggerSettings::setChannelAttenuation(double value)
...
 }
 
+void TriggerSettings::enableMixedSignalView()
+{
+	ui->extern_en->setDisabled(true);
+	ui->extern_to_en->setDisabled(true);
+	ui->mixedSignalLbl->setVisible(true);
+}
+
+void TriggerSettings::disaleMixedSignalView()
```
AlexandraTrifan, line 258: typo in the function name

AlexandraTrifan (approved):
Looks good!

src/trigger_settings.cpp
DanielGuramulta: fixed!

Issue #882, opened 2020-09-24
Pattern Generator: In number pattern, the color of the number set is always red after entering 


tagoylo:
**Environment:**
 - OS: Windows 10
 - FW Version: m2k-fw-v0.25-3-g1c38-dirty
 - SW Version or git tag: 7eec50d

**Describe the bug**
 In number pattern, the color of the number set is always red after entering a value although it is within the limit.

**To Reproduce**
1. Go to the pattern generator
2. Enable a channel, ex. Dio 0
3. Then set its pattern to Number. By default, the number is 0 and the color is not in red.
4. Type 0 again on the number box, now the color of 0 is red.

**Expected behavior**
Only red when the number is above the max number of grouped channels' capacity.

**Screenshots**
![image](https://user-images.githubusercontent.com/25096080/94146027-f3b4e900-fea5-11ea-9e0c-fd74968adce6.png)

DanielGuramulta:
Fixed with: 8fb53d4c46163a20faf9a831c31d522a48583efa

tagoylo:
This bug is fixed. Closing this item.

Issue #881, opened 2020-09-24
Pattern Generator: Scopy crashes when changing clock frequency (fine interval enabled)


tagoylo:
**Environment:**
 - OS: Windows 10
 - FW Version: m2k-fw-v0.25-3-g1c38-dirty
 - SW Version or git tag: 7eec50d

**Describe the bug**
Scopy crashes when changing the frequency down using the (-) button and fine interval is enabled (knob is with orange dot). This happens when at least two channels are enabled and pattern is selected.

**To Reproduce**
1. Enable at least 2 Dio channels.
2. Select Clock pattern for both channels.
3. Go to first channel and enable fine interval (with orange dot) for the clock frequency.
4. Then try increasing and then decreasing the frequency using the +/- buttons.
5. It is observed that after a few clicks on (-), Scopy lags and then crashes.

**Expected behavior**
Scopy should not crash.

**Screenshots**

![image](https://user-images.githubusercontent.com/25096080/94120675-fc47f800-fe82-11ea-884f-2153879247e5.png)

**Additional context**

[ini.bak and dmp files.zip](https://github.com/analogdevicesinc/scopy/files/5274536/ini.bak.and.dmp.files.zip)

DanielGuramulta:
Fixed with: d77bc7525dd04a060d6be9861f233c43c2ad1117

tagoylo:
This bug is fixed. Closing this item.

Issue #880, opened 2020-09-24
Pattern Generator: UART decoder does not display correct data


tagoylo:
**Environment:**
 - OS: Windows 10
 - FW Version: m2k-fw-v0.25-3-g1c38-dirty
 - SW Version or git tag: 7eec50d6

**Describe the bug**
UART decoder does not display correct data on the plot.

**To Reproduce**
1. Go to pattern generator
2. Enable a channel, ex DIO 0
3. Double click on it on the plot to open its channel settings.
4. Set pattern to UART
5. Set data, for example 1111
6. It does not show the data set on the plot. 

**Screenshots**
![image](https://user-images.githubusercontent.com/25096080/94111272-660dd500-fe76-11ea-912a-d750a38209fd.png)

DanielGuramulta:
Fixed with: 9dc1debcc7ff56302d8002ee9b5647a84d0a1087

tagoylo:
This bug is fixed. Closing this item.

Issue #877, opened 2020-09-23
Pattern Generator: Clicking Single button has no response


tagoylo:
**Environment:**
 - OS: Windows 10
 - FW Version: m2k-fw-v0.25-3-g1c38-dirty
 - SW Version or git tag: 7c68922

**Describe the bug**
When clicking Single, there is no response in the instrument.

**To Reproduce**
1. Open pattern generator
2. Enable a channel, Dio 0
3. Then double click on Dio 0 on the plot and change pattern to clock
4. Then click Single
5. There is no response on the instrument

**Expected behavior**
When clicking Single, the color of the button should momentarily in darker orange and the white box beside the instrument name should momentarily turn to a green triangle indicating it ran for a moment.

DanielGuramulta:
Fixed with: 0a601107ef19f5f8e69f77efdf2601c0996f3e7f

tagoylo:
This bug is fixed. Closing this item.

PR #883, opened 2020-09-24
Release fixes

DanielGuramulta:
Fixes for the following bugs

- #875 
- #876 
- #877 
- #880 
- #881 
- #882

Also the calibration thread was playing with the GUI sometimes causing weird crashes while drawing and polishing widgets. GUI stuff should be only handled from the main thread. Added a fix with: b98dc15.


AlexandraTrifan (approved):
Looks good

Issue #884, opened 2020-09-25
Pattern Generator: In binary counter pattern, the parallel items do not correspond to the timing diagram/plot


tagoylo:
**Environment:**
 - OS: Windows 10
 - FW Version: m2k-fw-v0.25-3-g1c38-dirty
 - SW Version or git tag: 7eec50d

**Describe the bug**
In binary counter pattern, the parallel items do not correspond to the timing diagram/plot. It also does not correspond to the number of channels grouped.

**To Reproduce**
1. Go to pattern generator
2. Enable at least 2 channels, Dio 0 and Dio 1
3. Then group the 2 channels.
4. Select group pattern as binary counter
5. Observe how the parallel items plot only has 0 displayed instead of having 0, 1, 2, 3. It does not correspond. It's always like a bit/s behind.
See screenshots for reference.

**Screenshots**
![image](https://user-images.githubusercontent.com/25096080/94231228-260e2700-ff36-11ea-8fac-ecc129526659.png)
![image](https://user-images.githubusercontent.com/25096080/94231310-5229a800-ff36-11ea-97fb-ab8cce049b15.png)

DanielGuramulta:
Fixed with: afb62be43fd86badff4f998aa596433cea0eb14e

tagoylo:
This bug is fixed. Closing this item.

PR #885, opened 2020-09-25
General Fixes

DanielGuramulta:
- Fix for: #884 
- Logic channels where not initially positioned in the right place.
- The scroll in the LA/PG was only moving "DIO0"

Signed-off-by: DanielGuramulta <Daniel.Guramulta@analog.com>


DanielGuramulta:
Added more fixes to this branch, renamed PR title.

AlexandraTrifan (approved):

PR #888, opened 2020-09-25
Fix issues related to exception handling while unplugging the board.

AlexandraTrifan:
- Unplugging the board during the calibration should not crash the app, even if the calibration is throwing exceptions. We handle those and disconnect without errors.
- Unplugging the board during the tool construction should not crash the app, even if the constructors are throwing exceptions. We handle those in the Tool Launcher and disconnect without errors.
- We should not suppose that the calibration and connection were done well and all the tools are available from that point on.

- Make sure the Power Supplies are not ON when first connecting to a board.
- Don't use averaging in the "Sample" Type for Spectrum Analyzer.

### DIFF
--- a/src/tool_launcher.hpp
+++ b/src/tool_launcher.hpp
@@ -260,6 +260,8 @@ private:
 
 	bool m_useNativeDialogs;
 
+	bool m_dac_tools_failed, m_adc_tools_failed;
+
 	void _setupToolMenu();
 	void saveRunningToolsBeforeCalibration();
 	void stopToolsBeforeCalibration();
--- a/src/toolmenuitem.cpp
+++ b/src/toolmenuitem.cpp
@@ -121,6 +121,15 @@ bool ToolMenuItem::eventFilter(QObject *watched, QEvent *event)
 	return QObject::event(event);
 }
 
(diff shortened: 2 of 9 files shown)


src/tool_launcher.hpp
```diff
@@ -260,6 +260,8 @@ private Q_SLOTS:
 
 	bool m_useNativeDialogs;
 
+	bool m_dac_tools_failed, m_adc_tools_failed;
```
DanielGuramulta, line 263: are this values initialised in the constructor of the ToolLauncher?

src/toolmenuitem.cpp
```diff
@@ -121,6 +121,15 @@ bool ToolMenuItem::eventFilter(QObject *watched, QEvent *event)
 	return QObject::event(event);
 }
 
+void ToolMenuItem::setCalibrating(bool calibrating)
```
DanielGuramulta, line 124: 👍

DanielGuramulta (approved):
Looks good!

src/tool_launcher.hpp
AlexandraTrifan: I added the initialization now.

PR #908, opened 2020-10-05
Preferences: initialize sig_gen_nr_of_periods to 1

adisuciu:
I missed this initialization when implementing https://github.com/analogdevicesinc/scopy/pull/833


DanielGuramulta (approved):
How does this affect other waveforms like sine square, where there was a minimum of 2 periods before?

adisuciu:
This only affects signal preview, so for sine/square/etc, the previewed waveform will have only one period.

DanielGuramulta:
> This only affects signal preview, so for sine/square/etc, the previewed waveform will have only one period.

Thx for the clarification.

Issue #900, opened 2020-09-30
Spectrum Analyzer: Importing saved Ref channel shows no name when selecting channels


tagoylo:
**Environment:**
 - OS: Windows 10
 - FW Version: m2k-fw-v0.25-3-g1c38-dirty
 - SW Version or git tag: 7eec50d

**Describe the bug**
Importing saved Ref channel shows no name when selecting channels. See screenshot for reference.

**To Reproduce**
1. Create a Ref channel by using Snapshot.
2. Then export Ref channel using Export in the Ref channel settings.
3. Then import reference channel by clicking + beside channel 2.
4. Browse the saved file.
5. Then click on Channels dropdown menu to see available channels. It can be observed that there is no name shown.

I also attached the Ref csv file I've used.

**Screenshots**
![image](https://user-images.githubusercontent.com/25096080/94653953-f0e74780-032e-11eb-941e-6acb3644ffbf.png)

**Additional context**

[ref.zip](https://github.com/analogdevicesinc/scopy/files/5303667/ref.zip)

AlexandraTrifan:
This should be fixed after commit 6e2c6ae35c3722a67864cc0bbae630ea9b9594aa

tagoylo:
Importing REF channel now shows CH on the list. Issue fixed. Closing this item.

Issue #899, opened 2020-09-30
Spectrum Analyzer: Axis scale lines messed up after changing unit from V/√Hz to dBFS/dBu/dBV


tagoylo:
**Environment:**
 - OS: Windows 10
 - FW Version: m2k-fw-v0.25-3-g1c38-dirty
 - SW Version or git tag: 7eec50d

**Describe the bug**
When changing units, from V/√Hz to dBFS/dBu/dBV the graticule axis scale lines is messed up.

**To Reproduce**
1. Go to the spectrum analyzer
2. Go to sweep settings and change the unit to V/√Hz
3. Then change back to dBFS/dBu/dBV
4. Then observe how the axis lines are compressed at the bottom part of the plot.

**Screenshots**
![image](https://user-images.githubusercontent.com/25096080/94643010-698ed980-0318-11eb-929d-f339797034a5.png)

AlexandraTrifan:
This should be fixed after commit bc38e73db07697b042f474819f6e3adf4275417b

tagoylo:
This bug is fixed. Closing this item.

PR #909, opened 2020-10-05
Release candidate: Fixes for the Spectrum analyzer

AlexandraTrifan:
- Fix bug #899 with commit 13bec282608113233a538c6ce1cf02a3c26a311f
- Fix bug #900 with commit e78efb42dd1d46713160e4ccffe1830e9e0572e1
- Allow the Frequency to be zero


src/spectrum_analyzer.cpp
```diff
@@ -34,6 +34,7 @@
 #include <QFileDialog>
 #include <QCheckBox>
 #include <QTimer>
+#include <QThread>
```
DanielGuramulta, line 37: is this include needed?

DanielGuramulta (approved):
Looks good!

src/spectrum_analyzer.cpp
AlexandraTrifan: removed

PR #915, opened 2020-10-07
TriggerSettings: Fix normal mode.

teoperisanu:
Bug description: switching between auto and normal mode, the normal mode would eventually result to behave as the auto mode.


DanielGuramulta (approved):
Looks good!

PR #920, opened 2020-10-08
Small fixes

adisuciu:


DanielGuramulta:
The siggen nr of periods is already on master f879e010da9619dbbb3a7a697f32a8123b40af28. Remove the duplicate commit

src/network_analyzer.cpp
```diff
@@ -1178,6 +1179,7 @@ void NetworkAnalyzer::goertzel()
 		}
 	}
 
+	Q_EMIT sweepStart();
```
DanielGuramulta, line 1182: Is this sweepStart signal used anywhere?

src/network_analyzer.cpp
```diff
@@ -702,6 +702,7 @@ NetworkAnalyzer::NetworkAnalyzer(struct iio_context *ctx, Filter *filt,
 
 	});
 
+	connect(this,SIGNAL(sweepDone()), ui->xygraph ,SLOT(reset()));
```
DanielGuramulta, line 705: 3 different styles for separating with `,` used in one line. `a,b`, `a, b` and `a ,b`. Just use one. (prefered `a, b`)

DanielGuramulta:
There's also a merge commit in here, remove it and rebase this branch with master

src/network_analyzer.cpp
adisuciu: It is now :stuck_out_tongue_winking_eye:

src/network_analyzer.cpp
adisuciu: This was sloppy. Fixed.

adisuciu:
Rebased

DanielGuramulta (approved):
Looks good!

PR #923, opened 2020-10-12
appveyor: Use the old deps archive provided by scopy-mingw-build-deps

AlexandraTrifan:
Use the old deps archive provided by scopy-mingw-build-deps instead of trying to get packages from repo.msys2.org.

Some older packages might be removed from the official repo. Using this
change, we don't depend on that official repo to build our application.

Also, remove the fast_finish flag, we want to use the MacOS builds whether
the Mingw build fails or not.

- [x] Wait for the scopy-mingw-build-deps build to be done, in order to test Scopy. ( https://ci.appveyor.com/project/analogdevicesinc/scopy-mingw-build-deps/builds/35695184 )


DanielGuramulta (approved):
Looks good!

PR #924, opened 2020-10-13
Windows Installer: Drop the SCOPY_PYTHONPATH environment variable. 

AlexandraTrifan:
Setting the environment variable might raise lots of issues, depending on what 
rights are used to install Scopy and also what rights are used to run Scopy.
This can be different for any user, so we drop the env var and figure another
way to check for the needed python modules.

We no longer set the env var. Version v1.1.2 will be the last one that
knows how to remove the SCOPY_PYTHONPATH env var from the system at
uninstall. If not uninstalled, then the env var will remain and we need
to provide backwards compatibility. That is why we append it to the end
of the path.

If Scopy was installed using the official installer, it will always use
the hardcoded paths in applicationDirPath(), where the python3.X folder is.
If Scopy is built locally on Windows, the user will set the SCOPY_PYTHONPATH
env var manually and, since that one is appended at the end, the program will
still function properly.

- [x] Remove commit  25dd7e9 from this PR after the build is done and tested (the commit is part of another PR).


DanielGuramulta (approved):
Looks good!

PR #926, opened 2020-10-14
WindowsInstaller: Use "Tasks" to create checkbox that deletes Scopy.ini.

AlexandraTrifan:
The checkbox for a task is checked by default, so we don't need to specify
another flag.
Also, the Appveyor job no longer needs to export an empty Scopy.ini.


DanielGuramulta (approved):
Looks great! will give it a try when artefacts will be available on appveyor.

PR #938, opened 2020-10-23
Appveyor: Use mingw64/bin/windres as RC compiler for 64 bit.

AlexandraTrifan:
Appveyor build failed because the RC compiler detection failed.
This might be caused by the CMake package update (checked repo.msys2.org and the update was done 2 days ago, when the builds started failing).
CMake does not detect an appropriate compiler for RC, so we point it to the right path.


DanielGuramulta (approved):
Looks good!

Issue #940, opened 2020-10-26
Manual range setting in Scopy Voltmeter


damercer:
When using the Voltmeter tool in Scopy I don't see a clear indication of which input range the M2k is in. Sometimes it looks like it might be in high gain or low gain range based on the amount of noise / size of the voltage steps. Does Scopy do some kind of Auto-ranging in the Voltmeter tool? It would be nice it there was some sort of indicator that changed (color maybe?) depending on which range it is in.

As an enhancement having a manual override so that the user can turn off any auto-ranging and select which range is to be used. There could be an "out-of-range" warning if the voltage is at or beyond the range that is manually set.

Doug

adisuciu:
I just checked, the voltmeter in Scopy always uses low-gain. I can change it to have LOW/HIGH/AUTO .. 

-Adrian

damercer:
That explains why the readings were so much more noisy when measuring a small differential voltage ( < 500 mV ) compared to the mean value measured in the scope on the 0.5 V/div range.

That would be a good feature to enhance the Voltmeter usability for small DC voltages.

Thanks

Doug

adisuciu:
We implemented this here: https://github.com/analogdevicesinc/scopy/pull/942

You can give this a test if you want https://ci.appveyor.com/project/analogdevicesinc/scopy/builds/36108722/job/fhvldk7vygj16han/artifacts

It has high gain/low gain/ auto gain. Auto switches from low to high as soon as it detects an overflow, and from high to low after there have been 10 readings below a threshold.

-Adrian

PR #942, opened 2020-10-27
Voltmeter manual range

DanielGuramulta:
Implements feature described in #940

### DIFF
--- a/src/dmm.cpp
+++ b/src/dmm.cpp
@@ -33,6 +33,10 @@
 #include <gnuradio/blocks/skiphead.h>
 #include <gnuradio/blocks/delay.h>
 #include <gnuradio/filter/dc_blocker_ff.h>
+#include <gnuradio/blocks/max_blk.h>
+#include <gnuradio/blocks/min_blk.h>
+#include <gnuradio/blocks/stream_to_vector.h>
+#include <gnuradio/blocks/vector_to_stream.h>
 
 #include <boost/make_shared.hpp>
 
@@ -69,7 +73,10 @@ DMM::DMM(struct iio_context *ctx, Filter *filt, ToolMenuItem *toolMenuItem,
 	filename(""),
 	use_timer(false),
 	logging_refresh_rate(0),
-	wheelEventGuard(nullptr)
+	wheelEventGuard(nullptr),
+	m_autoGainEnabled({true, true}),
+	m_historyForGain({0, 0}),
+	m_gainHistorySize(10)
 {
 	ui->setupUi(this);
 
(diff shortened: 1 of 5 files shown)


src/dmm.cpp
```diff
@@ -267,6 +331,90 @@ bool DMM::isIioManagerStarted() const
...
+	const double lo_hi = lo + 0.1;
+	const double lo_lo = lo - 0.1;
+
+	auto insideDeltaZone = [&](double voltage) {
+		return (voltage > hi_lo && voltage < hi_hi)
+				|| (voltage < lo_hi && voltage > lo_lo);
+	};
+
+	// TODO: can we querry analog in channels count?
+	for (int i = 0; i < 2; ++i) {
```
AlexandraTrifan, line 358: Can't we use **m_adc_nb_channels** or **m_m2k_analogin->getNbChannels()** ?

DanielGuramulta: Thanks. Forgot about this methods.

src/dmm_api.cpp
```diff
@@ -189,4 +189,20 @@ void DMM_API::setNotes(QString str)
...
 
+QVector<int> DMM_API::getGainModes() const
+{
+	return {dmm->ui->gainCh1ComboBox->currentIndex(),
+		dmm->ui->gainCh2ComboBox->currentIndex()};
+}
+
+void DMM_API::setGainModes(const QVector<int> &gainModes)
+{
+	if (gainModes.size() < 2) {
```
AlexandraTrifan, line 200: We can also use **m_adc_nb_channels** here, instead of 2..right?

DanielGuramulta: Yup we can use it here, as for each channel a gain control is associated. I think this can be done in a better way by providing a `DMMChannelWidget`  thus having to instantiate one for each channel (in `m_adc_nb_channels`) instead of duplicating everything. Each having an API with individual controls, but that's not the purpose of this pr.

AlexandraTrifan (approved):

DanielGuramulta:
- Changed the algorithm to look at the min / max value in the input(max/min peak) rather than the processed displayed value.
- When going from Low gain to High gain we hold a history of size N. Only when N captures correspond to the High gain mode we change to it.
- Changing from High gain to Low gain is done instantaneous.

DanielGuramulta:
@AlexandraTrifan I have made some changes since the initial pull request, can you review this again?

AlexandraTrifan (approved):
Looks good. Will run a test later today

PR #952, opened 2020-11-03
resources: translations: scopy_de.ts: Add German translation

mhennerich:
Signed-off-by: Michael Hennerich <michael.hennerich@analog.com>

### DIFF
--- a/resources/translations/scopy_de.ts
+++ b/resources/translations/scopy_de.ts
@@ -1,0 +1,5967 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!DOCTYPE TS>
+<TS version="2.1" language="de_DE">
+<context>
+    <name>BaseMenu</name>
+    <message>
+        <location filename="../../ui/basemenu.ui" line="14"/>
+        <source>Form</source>
+        <translation>Form</translation>
+    </message>
+</context>
+<context>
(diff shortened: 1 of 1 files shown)


DanielGuramulta (approved):
A `scopy_de.qm` should also be added here (https://github.com/analogdevicesinc/scopy/blob/master/resources/translations.qrc). This will bundle the generated qm file as a resource, thus making it available for packaged Scopy (from windows installer, flatpak, ...)

mhennerich:
> A `scopy_de.qm` should also be added here (https://github.com/analogdevicesinc/scopy/blob/master/resources/translations.qrc). This will bundle the generated qm file as a resource, thus making it available for packaged Scopy (from windows installer, flatpak, ...)

hmm shouldn't this not been done automatically during build?

DanielGuramulta:
> > A `scopy_de.qm` should also be added here (https://github.com/analogdevicesinc/scopy/blob/master/resources/translations.qrc). This will bundle the generated qm file as a resource, thus making it available for packaged Scopy (from windows installer, flatpak, ...)
> 
> hmm shouldn't this not been done automatically during build?

Totally agree. I am not sure what where the limitations of generating the resources file, when this was implemented but I will have a look at how it can be done.

PR #954, opened 2020-11-05
Tool Launcher: Handle possible exceptions thrown from tools that don't require calibration.

AlexandraTrifan:
This handles the crashes that might happen if the board is unplugged during connection, right before starting the calibration.
This commit also changes a bit, the flow of the connection:
- We stop the Search Timer before creating the context. We create the context, handle any issues that might appear, without worrying that the Search Timer will remove an M2K from the list while we did not manage to uninit everything related to it.
- Start the Alive Timer after every tool was properly initialized and contructed, not after the iio_context creation. A lot of exceptions can raise between constructing the iio_context and finishing up all the Tools.
- Disconnect if anything failed during the "switchContext" phase (which created the iio_context and all the tools). The disconnect phase will restart the Search Timer and remove all the unplugged devices from our list. 

Signed-off-by: AlexandraTrifan <Alexandra.Trifan@analog.com>

### DIFF
--- a/src/tool_launcher.cpp
+++ b/src/tool_launcher.cpp
@@ -1048,9 +1048,12 @@ void adiscope::ToolLauncher::disconnect()
 		destroyContext();
 		loadToolTips(false);
 		resetStylesheets();
+		auto infoPg = selectedDev->infoPage();
+		if (infoPg) {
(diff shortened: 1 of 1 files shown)


DanielGuramulta (approved):
Looks good!

PR #956, opened 2020-11-09
Oscilloscope: prevent popup when creating instrument

adisuciu:
This happens because a label with no parent is set visible before
inserting it in a layout


teoperisanu (approved):

DanielGuramulta (approved):
Looks good!

Issue #955, opened 2020-11-09
Disconnect device is slow because of unneeded QSettings syncs


adisuciu:
https://github.com/analogdevicesinc/scopy/blob/master/src/spinbox_a.cpp#L134

This delete has a runtime of 3-400 ms because it always syncs the settings file (at least on Windows). Since there are a lot of spinboxes in Scopy, this causes significant delay on disconnect. Please investigate this behavior.

adisuciu:
https://github.com/analogdevicesinc/scopy/pull/957

PR #957, opened 2020-11-09
Settings saving performance

DanielGuramulta:
Should fix the issue described in: #955

@adisuciu can you test this with the same method you used to find out about this issue and see if there are any improvements?


DanielGuramulta:
The singletone wrapper should also have copy constructor/ copy assig. deleted. Will add this if this approach is correct for fixing the issue.

adisuciu (approved):

PR #959, opened 2020-11-09
InfoPage: Added libm2k version

adisuciu:

### DIFF
--- a/src/info_page.cpp
+++ b/src/info_page.cpp
@@ -412,6 +412,8 @@ M2kInfoPage::~M2kInfoPage()
 void M2kInfoPage::getDeviceInfo()
 {
 	InfoPage::getDeviceInfo();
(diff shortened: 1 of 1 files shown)


DanielGuramulta (commented):
Is the libm2k version in the info page related with the connected board (ex: m2k). For example the libiio version in the info page from what I remember is related to the libiio on the m2k, not the one with whom Scopy was built. Wouldn't it make more sense to have this libm2k version in an "about" page where we could also list other dependencies that are related to Scopy (Qt version, libiio, libusb etc)?

AlexandraTrifan:
I agree. We could have an "About" page on the Homepage, or in some place easily accessible. This page should show all the important compile/runtime dependencies and their version, while the device page shows everything related to the specific device.

PR #968, opened 2020-11-26
PatternGenerator: Fix i2c pattern crash that occurs when overflowing …

adisuciu:
…nr of samples


DanielGuramulta (approved):
Looks good!

PR #969, opened 2020-11-26
About page

adisuciu:

PR #970, opened 2020-11-26
Strip dlls and python

adisuciu:


DanielGuramulta (changes requested):
This pull request seems to be a duplicate of: #969. Separate the commits in each one so that you have one pull request / feature (about page, strip dlls)

adisuciu:
They depend on oneanother, when about_page will be merged, this will be only one commit.

DanielGuramulta (approved):

AlexandraTrifan (approved):

PR #973, opened 2020-12-02
PatternGenerator: Fix overflow in SPI and UART pattern generation

adisuciu:
This overflow occurs when the requested buffersize is smaller than the
requested/suggested buffersize by the pattern


DanielGuramulta (approved):
Looks good!

PR #974, opened 2020-12-03
Manual Calibration and Debugger fixes

AlexandraTrifan:
Manual Calibration: 
- Fix the file dialog used for calibration saving.
- Make sure the tool does not use the calibration coefficients in the calibration process (since this will alter and output wrong calib coefficients).

Debugger:
- Fix the way attributes are loaded (whether they have available values or not).

The calibration coefficients will properly work in Scopy only after the corresponding libm2k PR is merged ( https://github.com/analogdevicesinc/libm2k/pull/207 ) and the scopy-mingw-build-deps is rebuilt to contain the new version.

### DIFF
--- a/src/debug.cpp
+++ b/src/debug.cpp
@@ -147,7 +147,7 @@ void Debug::scanChannels(const QString& devName)
 void Debug::scanChannelAttributes(QString devName, QString& channel)
 {
 	struct iio_device *device;
-	struct iio_channel *ch;
+	struct iio_channel *ch = nullptr;
 	const char *attr;
 	unsigned int nb_attrs;
 	bool isOutput;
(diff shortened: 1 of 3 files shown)


src/debug.cpp
```diff
@@ -166,36 +166,37 @@ void Debug::scanChannelAttributes(QString devName, QString& channel)
 		attributeAvailable.clear();
 		filename.clear();
 
-		if (channel != QString("Global")) {
+		bool isGlobal = (channel == QString("Global"));
```
DanielGuramulta, line 169: this can be const

DanielGuramulta (approved):
Looks good!

PR #1002, opened 2021-02-12
MetricFormatter: optionally remove trailing zeroes from multiprecisio…

adisuciu:
…n variables

### DIFF
--- a/src/plot_utils.cpp
+++ b/src/plot_utils.cpp
@@ -54,6 +54,17 @@ bool PrefixFormatter::getTwoDecimalMode()
 
 QString PrefixFormatter::buildString(double value, QString prefix, QString unitType, int precision) const
 {
+
+	if (m_trimZeroes) {
+		for (auto i = 0; i < precision; i++) {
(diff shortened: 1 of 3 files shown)


src/plot_utils.cpp
```diff
@@ -54,6 +54,16 @@ bool PrefixFormatter::getTwoDecimalMode()
 
 QString PrefixFormatter::buildString(double value, QString prefix, QString unitType, int precision) const
 {
+
+	if(m_trimZeroes) {
```
DanielGuramulta, line 58: nitpick: missing space `if ()`
also after `for (`
`if (single...`

src/plot_utils.cpp
```diff
@@ -54,6 +54,16 @@ bool PrefixFormatter::getTwoDecimalMode()
...
+		for(auto i = 0; i < precision; i++) {
+			double singlePrecision = value * pow(10,i);
```
DanielGuramulta, line 60: this can be `const`

DanielGuramulta (approved):
Looks good!

Issue #823, opened 2020-07-14
Connecting to virtual devices


balping:
Is it possible to use Scopy to connect to a virtual device, ie a software emulated oscilloscope that is supplying generated real-time raw acquisitions? I'm willing to modify the source if this is possible in theory. I looked a bit into the source and I have the feeling that it should be possible to emulate an iio device and communicate with Scopy. Is this far fetched? Has anyone done something similar?

rgetz:
@balping - I on't think anyone has looked at this.

Is it better to use a virtual device? or an audio sink/source?

rgetz:
Eventually - we need to be able to support audio in/out (soundcard)

https://www.codeproject.com/Articles/421287/Cross-Platform-Microphone-Audio-Processing-Utility

adisuciu:
ADALM2000 emulator implemented .. soundcard support should be a task for iio-emu

PR #1006, opened 2021-02-18
Add ADALM2000 emulator.

teoperisanu:
In order to emulate the behavior of ADALM2000 , Scopy launches a server (iio-emu) that can communicate via network backends with libiio clients. The server depends on libtinyiiod for the iiod network protocol.

https://github.com/analogdevicesinc/iio-emu
https://github.com/analogdevicesinc/libtinyiiod/

Related to #823.


ui/connect.ui
```diff
@@ -119,6 +119,57 @@ color: green;
...
+        <string notr="true">QPushButton {
+  min-height: 30px;
+  max-height: 30px;
+  border: 0px;
+}</string>
+       </property>
+       <property name="text">
+        <string>Enable Demo</string>
+       </property>
+       <property name="blue_button" stdset="0">
```
DanielGuramulta, line 168: 👍

src/connectDialog.cpp
```diff
@@ -50,6 +51,10 @@ ConnectDialog::ConnectDialog(QWidget *widget) : QWidget(widget),
 	setDynamicProperty(ui->hostname, "invalid", false);
 	setDynamicProperty(ui->hostname, "valid", false);
 	ui->infoSection->hide();
+
+	ui->demoDevicesComboBox->addItem("ADALM2000");
+	ui->demoDevicesComboBox->setDisabled(true);
+	process = std::unique_ptr<QProcess>(new QProcess(this));
```
DanielGuramulta, line 57: Does passing `this` as parent to the `QProcess` not ensure that the object will be destroyed when `this` gets destroyed? If yes then having it stored also in unique_ptr would result in a double delete. Either remove `this` as parent or the unique_ptr.

src/connectDialog.cpp
```diff
@@ -86,10 +91,54 @@ void ConnectDialog::btnClicked()
...
+				process->setProgram(program);
+				process->start();
+				started = process->waitForStarted();
+				if (!started) {
+					qDebug() << "Process failed to start";
+				}
+			}
+			if (started) {
+				qDebug() << "Process started successfully";
+				QThread::msleep(100);
```
DanielGuramulta, line 115: Why do we need this sleep here? It currently blocks the main thread for `100ms`

DanielGuramulta (approved):
Looks good, should `iio-emu` be reviewed with this pull request?

src/connectDialog.cpp
teoperisanu: This is a simple way to ensure that the server is prepare to listen to the clients, otherwise the validateInput function cannot create a context.
There is also the possibility to send a message when the server is ready.

src/connectDialog.cpp
teoperisanu: Modified.

PR #1013, opened 2021-02-24
Azure pipelines clang format

DanielGuramulta:
This adds an initial azure pipelines build that checks if the code is properly formatted (as .clang-format wants it to be).
If the code is formatted ✅ , else the build will fail stating the code is not properly formatted.
Also a patch will be created that can be applied to fix the issue where clang-format on the developers system is not installed or not available. The following message is also shown:
```
The code style issues can be fixed by applying the format-code.patch found in the artifacts sections
The command to apply the patch is: git apply format-code.patch
```

The formatting job is based on an `ubuntu-20.04` image with `clang-format-11` installed and used.

Currently the build will fail as the code is not formatted. This will be done after the next release is published:

- [ ] rebase pr after the new release is published
- [ ] format the code base

[EDIT]: As an example, for the current failing build this is the patch generated that can format the code: https://dev.azure.com/AnalogDevices/M2k/_build/results?buildId=1597&view=artifacts&pathAsName=false&type=publishedArtifacts


DanielGuramulta:
This feature was added for the branch scopy-2.0. Will close this

PR #1014, opened 2021-02-25
Small fixes

ioanachelaru:


adisuciu:
image
https://github.com/analogdevicesinc/scopy/blob/master/src/signal_generator.cpp#L1085 - should be updated to +/-5V

adisuciu (approved):

PR #1017, opened 2021-02-26
phonehome: fix phonehome firmware label

adisuciu:
The phonehome firmware label was updated before actually getting firmware
information from the board, causing it to always return firmware update
required

### DIFF
--- a/src/info_page.cpp
+++ b/src/info_page.cpp
@@ -63,20 +63,6 @@ InfoPage::InfoPage(QString uri, Preferences *pref, PhoneHome* phoneHome,
 		this, SLOT(ledTimeout()));
 	connect(m_blink_timer, SIGNAL(timeout()),
 		this, SLOT(blinkTimeout()));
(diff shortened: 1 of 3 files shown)


src/info_page.cpp
```diff
@@ -249,6 +235,23 @@ void InfoPage::refreshInfoWidget()
 	else
 		setStatusLabel(ui->lblIdentifyStatus, tr("Your hardware revision does not support the identify feature"));
 
+	if(m_phoneHome->getDone())
+	{
+		const int checked = dynamic_cast<M2kInfoPage*>(this)->checkLatestFwVersion(m_info_params.value("Firmware version"));
+		if (checked == 1) {
+			ui->lblFirmware->setText("Firmware is up to date!");
```
DanielGuramulta, line 242: Should this text be wrapped in a `tr()` for translations?

adisuciu: fixed

teoperisanu (approved):

PR #1044, opened 2021-03-12
Scopy 2.0 custom widgets

ioanachelaru:


DanielGuramulta (approved):
PR looks good, clang-format and cmake-format must be applied to some files

PR #1048, opened 2021-03-15
Theme Management: Implemented ThemeManager logic

ioanachelaru:
Signed-off-by: Ioana Chelaru <Ioana.Chelaru@analog.com>


gui/src/theme_manager.cpp
```diff
@@ -0,0 +1,39 @@
+#include <QFile>
+#include <QIcon>
+
+#include <scopy/gui/theme_manager.hpp>
+
+using namespace scopy::gui;
+
+ThemeManager::ThemeManager()
```
DanielGuramulta, line 8: should initialise here in the `initializer list` m_app to `nullptr`

gui/src/theme_manager.cpp
```diff
@@ -0,0 +1,39 @@
...
+{
+	m_availableThemes = new QStringList();
```
DanielGuramulta, line 10: this is not required to be allocated with new, `QStringList m_availableThemes`

gui/src/theme_manager.cpp
```diff
@@ -0,0 +1,39 @@
...
+
+void ThemeManager::setCurrentTheme(const QString& theme)
+{
+	m_currentTheme = theme;
+
+	QFile file(":/stylesheets/" + theme + ".qss");
+	file.open(QFile::ReadOnly);
+	QString stylesheet = QString::fromLatin1(file.readAll());
+
+	m_app->setStyleSheet(stylesheet);
```
DanielGuramulta, line 32: What happens here if `m_app` was not set? Should check if m_app is nullptr

gui/src/theme_manager.cpp
```diff
@@ -0,0 +1,39 @@
...
+	return INSTANCE;
+}
+
+void ThemeManager::setApplication(QApplication* app) { m_app = app; }
+
+void ThemeManager::setCurrentTheme(const QString& theme)
+{
+	m_currentTheme = theme;
+
+	QFile file(":/stylesheets/" + theme + ".qss");
```
DanielGuramulta, line 28: We have this kind of read from file logic in a `utils` class. If that was not yet moved add a `TODO` here to use the utils function to avoid duplicate code.

gui/include/scopy/gui/theme_manager.hpp
```diff
@@ -0,0 +1,32 @@
...
+{
+public:
+	static ThemeManager& getInstance();
+
+	ThemeManager(const ThemeManager&) = delete;
+	ThemeManager& operator=(const ThemeManager&) = delete;
+
+	void setApplication(QApplication* app);
+	void setCurrentTheme(const QString& theme);
+	QString getCurrentTheme();
```
DanielGuramulta, line 19: this get methods could be const

DanielGuramulta (approved):

PR #1052, opened 2021-03-18
Scopy 2.0: Added UserNotes, InstrumentNotes and InfoWidget

ioanachelaru:


gui/include/scopy/gui/user_notes.hpp
```diff
@@ -0,0 +1,127 @@
...
+#include <QButtonGroup>
+#include <QPushButton>
+#include <QSettings>
+#include <QTextBrowser>
+#include <QWidget>
+
+#include <scopy/core/api_object.hpp>
+#include <vector>
+
+using namespace std;
```
DanielGuramulta, line 33: This is considered bad practice bringing the whole std namespace, you could just use `std::vector` or instead of `using namespace std;` just `using std::vector;`

gui/include/scopy/gui/user_notes.hpp
```diff
@@ -0,0 +1,127 @@
...
+{
+	friend class UserNotes_API;
+
+	Q_OBJECT
+
+public:
+	explicit UserNotes(QWidget* parent = 0);
+	~UserNotes();
+
+	scopy::core::ApiObject* m_api();
```
DanielGuramulta, line 58: namig should be `getApi` or something like that w/o `m_`

gui/src/info_widget.cpp
```diff
@@ -0,0 +1,60 @@
...
+	m_timer = new QTimer(this);
+	connect(m_timer, SIGNAL(timeout()), this, SLOT(updatePosition()));
+	this->hide();
+}
+
+InfoWidget::~InfoWidget() { delete m_ui; }
+
+void InfoWidget::updatePosition()
+{
+	QPoint p = this->parentWidget()->mapFromGlobal(QCursor::pos());
```
DanielGuramulta, line 42: this can be const `const QPoint p`

gui/src/info_widget.cpp
```diff
@@ -0,0 +1,60 @@
...
+
+	// Position the widget in the bottom right corner of the pixmap
+	// that is dragged
+	this->move(p.x() + 159, p.y() + 54);
+}
+
+void InfoWidget::setText(QString text) { m_ui->lbl->setText(text); }
```
DanielGuramulta, line 49: string should be passed as `const QString &text`

gui/src/instrument_notes.cpp
```diff
@@ -0,0 +1,21 @@
...
+
+	m_ui->widgetSubsectionSeparator->setLabel("Note");
+	m_ui->widgetSubsectionSeparator->setButton(false);
+}
+
+InstrumentNotes::~InstrumentNotes() { delete m_ui; }
+
+QString InstrumentNotes::getNotes() { return m_ui->textEdit->toPlainText(); }
+
+void InstrumentNotes::setNotes(QString str) { m_ui->textEdit->setPlainText(str); }
```
DanielGuramulta, line 21: string should be passed as `const QString &str`

PR #1061, opened 2021-03-23
Scopy 2.0 custom widgets

ioanachelaru:


gui/src/application_restarter.cpp
```diff
@@ -0,0 +1,31 @@
+#include "application_restarter.hpp"
+
+#include <QApplication>
+#include <QDir>
+#include <QProcess>
+
+using namespace scopy::gui;
+
+ApplicationRestarter::ApplicationRestarter(const QString& executable)
```
DanielGuramulta, line 9: I think that the application restarter should be in scopy::core as its not a GUI related class

DanielGuramulta (approved):
Looks good, https://dev.azure.com/AnalogDevices/M2k/_build/results?buildId=2215&view=results some code still has some styling issues, could this issues also be fixed in this pr so that all the checks would pass. Thanks

PR #1062, opened 2021-03-24
PatternGenerator: Set min sampling frequency to 1kHz.

teoperisanu:
This change will avoid creation of tiny buffers.
Number Pattern uses this function.

Signed-off-by: Teo Perisanu <Teo.Perisanu@analog.com>


DanielGuramulta (approved):
Looks good!

PR #1113, opened 2021-04-20
V130 rc1 fixes

adisuciu:


AlexandraTrifan (approved):
There is some commented code. Can that be removed?
Looks good otherwise.

PR #1118, opened 2021-04-27
Created separate menus for all tools

ioanachelaru:


AlexandraTrifan (approved):
Looks good.
But I need to mention 2 things here:
- the gui/trigger_settings.cpp has an extra # include to a "scopy_exception_handler" (probably will be used later, but should now be removed/commented) 
- the core/CMakeLists.txt seems to be missing a link to libiio, but that is not related to the topic of this PR.

PR #1129, opened 2021-06-10
Disable libiio DNS, AIO and USB backend for ubuntu builds.

teoperisanu:
DNS requires libavahi.
AIO requires libaio.
USB backend requires libusb.

Because these are just test builds, disable the options instead of installing the dependencies.


teoperisanu:
Already fixed in #1127.

PR #1174, opened 2021-11-08
android: fix gnuradio in external storage bug

adisuciu:
Previously the application crashes if there is no gnuradio folder in
external storage root. This fixes it by setting an envvar to the cache
of the application.


ioanachelaru (approved):

adisuciu:
Docker rebuild is also required

PR #1185, opened 2021-11-23
Dev probe attenuation

IonutMuthi:
Updated Probe Attenuation values to include : 0.2; 0.5; 2; 5; 20; 50; 200; 500


adisuciu:
Squash this to a single commit, remove useless bool and then we can merge.

adisuciu (approved):

PR #1198, opened 2022-01-20
plots: plot optimizations

adisuciu:
- [x] fix android
- [ ] fix CI (build correct version of qwt on all platforms)


nanoant:
Hi @adisuciu I finally managed to test your changes. And they look pretty encouraging. While enabling OpenGL seems to do good in general, but removing 10 FPS cap in favor of 60 FPS actually shows that there are pretty significant bottlenecks.

First, I commented out `#define IMMEDIATE_PAINT` as it was causing pretty excessive CPU usage, mostly in Qwt rendering function and also causing not really smooth rendering.

Once commented out I still have around 140-150% (1.4-1.5 core) usage on my six core i7-8700B but solid 160 FPS on oscilloscope. That looks really nice now. (You should think of digital phosphor function now)

NOTE: Of course I understand 160 FPS is theoretical out of render ms time, since the both rendering is still capped to 60 FPS V-sync and readout is also capped at 60 FPS.

Zrzut ekranu 2022-02-2 o 22 09 38

Back to performance, I ran profiler and found out that only 25% of that is spend for rendering now, rest is essentially libm2k, either main two threads and many many `gr:thread::bodywrapper`. I build everything with `CMAKE_BUILD_TYPE=RelWithDebInfo` so not sure if such a high CPU usage is expected or what is the culprit here.

Zrzut ekranu 2022-02-2 o 23 01 01
Zrzut ekranu 2022-02-2 o 23 01 15
Zrzut ekranu 2022-02-2 o 23 01 20
Zrzut ekranu 2022-02-2 o 23 01 24
Zrzut ekranu 2022-02-2 o 23 01 33
Zrzut ekranu 2022-02-2 o 23 01 58

Best regards! Adam.

adisuciu:
We're using gnuradio for our signal chain and gr::thread::thread_body_wrapper is probably related to that. We will do some additional profiling on libm2k when we get a chance.
We also need to review our signal chains as some optimizations may be due there as well. We will probably do that when we upgrade to gnuradio 3.10 some time in the future.

Thank you for your valuable insight ! I would've given up on OpenGL rendering if it weren't for your comment :) 

-Adrian

adisuciu:
Okay, I recompiled without IMMEDIATE_PAINT and i can see some stuttering when interacting with the UI which suggests that something more is going on in the UI thread than it should. I'll investigate.

nanoant:
@adisuciu I need to confess that I have started writing today some smaller (minimal deps) and quicker (Qt + QtCharts) m2k app to see if I can get somehow decent CPU usage. Here's what I managed to achieve so far: https://github.com/nanoant/cutescope

I have no intention to compete with Scopy, but I need something that has minimal CPU impact and minimal natively looking UI. And so far I get solid 60 FPS without causing my Mac mini fans spinning.

adisuciu:
No worries, I tested your build though, your CPU usage is not that much better if you do 8k samples 10MSPS. However do you get fans spinning if you use these parameters ? And another question, did you have the same fan spinning issue before we started testing this OpenGL stuff ?

I think I messed up removing the frame cap - [`f673081` (#1198)]([`f673081` (#1198)](https://github.com/analogdevicesinc/scopy/pull/1198/commits/f67308169166ef0f4b3c88859de86ea5b5b93a17#diff-570ac94c1fc3aba434fbaabe697a51701584853a767aecb29fa9d4e8fcd1040dL124-L125) , https://github.com/analogdevicesinc/scopy/pull/1198/commits/f67308169166ef0f4b3c88859de86ea5b5b93a17#diff-6af1d1404c9e556e02e619af74a8f13f0717f8808578952250107e6032465d5dL121-L122) - these should've been updated as well 
My guess is that fans start going off because there was no frame cap on XY and histo plots so the gpu was going crazy. The histo is computed even when you don't see it because it is used for measurements. 

We're actually acquiring more data and throwing samples away if we're out of render time. So data acquisition always goes at the maximum possible rate - that is what the gr:: threads are. I'm looking into a way to rate limit the acquisition so it goes at the same speed as the plotting.

I force pushed again - with IMMEDIATE_PAINT commented out and fps cap patched, if you're interested in testing, the CPU usage should be about the same but I'm curious if it's more usable now and/or if the fans calmed down. 

-Adrian

adisuciu:
With some additional optimizations in libm2k and gr-m2k i brought cpu usage down to 40%-50% at 60fps .. I'll clean up the changes and push them this week. You will need to rebuild the dependencies. I'll let you know when they are up.

nanoant:
Great! Definitely looking forward to it. Ahh... one more thing. I am using 4K display, therefore this explains why the rendering may be CPU/iGPU intensive.

Btw. do you have vector sources for your icon assets? It would be nice to replace them all with SVGs as low-res icons don't look great on Retina/Hi-DPI screen. Also then you'd not need to make distinction for light and dark theme for icons as all SVG icons could be made monochromatic.

I also remade macOS icon if you are interested:
Scopy icon

adisuciu:
Hello, we unfortuantely have only part of the icons in vector format :( 

I pushed to this branch and to https://github.com/analogdevicesinc/gr-m2k . There is still one corner case left I can't figure out where if you set maximum number of samples (by changing resolution bandwidth) and then forcing some addtitional channels by snapshotting, the application freezes because of high CPU usage on the GUI thread. It's not related to rendering but rather to the huge number of points. 

A possible workaround is to lower the FPS in those cases but I want to put in a more general solution if possible. I'll think about it some more. You can test if you want. It should be smooth in 95% of the cases.

nanoant:
> A possible workaround is to lower the FPS in those cases but I want to put in a more general solution if possible. I'll think about it some more. You can test if you want. It should be smooth in 95% of the cases.

Capping the acquisition speed works extremely well. With 60 FPS target my Mac mini uses 70% CPU (100% = single core) and stays at ~75 C deg. With 30 FPS target only 35% CPU is used and ~65 C deg temperature. Well done @adisuciu!

I think having target FPS setting exposed (not hidden) e.g. as a 60, 30, 15, 10, 5 combo would be great. Especially if you run on laptop/tablet and you want to trade battery for less smooth operation. Also a note of higher FPS drains battery faster would be useful then.

Anyway I guess this is a great improvement to what was there so far hard-capped at 10 FPS and software only rendering.

> Hello, we unfortuantely have only part of the icons in vector format. I can ask the designer if he still has them ? or ever had them ?

If you ask the designer what is still available as vector then I could rework raster-only files into SVG keeping the pixel grid and 1x output intact. I did that many times in the past and it is fun after hours work. I could work on that and send PR for turning all icons into SVG. I did that already for quite a few projects in the past.

nanoant:
@adisuciu I just want to thank you again for all the recent changes in `port-qwt`. These are absolutely great improvements. Scopy is now running great at 30FPS cap with my CPU being cool and 4K rendering being smooth and great looking with OpenGL on. It is like completely different class of software now. I believe  all other users will appreciate these changes too!

image

image

adisuciu:
I'm glad to hear that. This plot optimization quest was very interesting. It started out as very basic optimizations, I tried porting to OpenGL but with the previous version of qwt it failed miserably. When I tried porting to the new qwt-multiaxes I came about a lot of errors and had no idea what the effort would be so I dropped it. 
After talking to you I realized that the effort is small and the benefit is huge. So thanks for that ! 

There are a few things that I still need to do.
Cap to 30FPS by default and fix the CI 
Linux - works
Android - works
Windows - WIP
macOS - would work but takes too long to build for some reason - will investigate

As for the SVG icons I'm still waiting on those ...

Anyways I'll merge this PR pretty soon. Thanks for all your help :)

pcercuei:
@adisuciu by the way, back in ~2015 or so we implemented an algorithm in GtkDataBox to boost the rendering of plots on the iio-oscilloscope. It works by shrinking the amount of data to fit the plot. For instance if your buffer is 4096 samples and your plot is 100 pixels wide, it will shrink the buffer to 100 samples, and the output would still look virtually the same, but the time spent plotting is significantly reduced.

Qwt might or might not be doing something similar internally, though.

For reference, the GtkDataBox patch in question: https://github.com/analogdevicesinc/iio-osc-mingw/blob/master/gtkdatabox-0.9.2.0.patch

adisuciu:
Paul, I think this is what you are reffering to -  https://qwt.sourceforge.io/class_qwt_plot_curve.html#a96db1b854c63bfbc452c943251a11b66a4aed183ef24167cb6df280abf166cb7b .

I added this filtering to all curves in the FFT plot - https://github.com/analogdevicesinc/scopy/blob/port-qwt/src/FftDisplayPlot.cc#L94 - although I didn't play enough with these to see how much of a difference it makes.

Problem was that especially on high DPI displays Qt does an internal rescaling which really hits performance when not using hardware acceleration.

Another problem, which we haven't solved yet - is when plotting a very high number of points (>1M), the algorithm to filter out the points still takes a super long time.

There is also a QwtCurveFitter - https://qwt.sourceforge.io/class_qwt_plot_curve.html#ac15588c78d175906a30de501b4dd7957 but I don't think this is something that we want (?)

-Adrian

adisuciu:
@nanoant I have the svg sources .. unfortunately the png and svg file names do not match but they should all be there ..
[svg_sketch (1).zip](https://github.com/analogdevicesinc/scopy/files/8160447/svg_sketch.1.zip)

If you want to change the sources from png to svg, I'll gladly accept your PR.

-Adrian

nanoant:
> If you want to change the sources from png to svg, I'll gladly accept your PR.

@adisuciu Hi. Thanks for the files. I have created PR #1215, it was quite challenging task, since there were plenty of unused files. Scopy looks much better on my screen now :) There are still some things to do, but this PR sets a foundation for HiDPI scalable assets and deprecated most of .png. But let's discuss this in #1215

adisuciu:
Thank you so much for your pull requests! I know things are pretty messy in the resources folder. :) We'll review the pull requests and merge them as soon as possible so they are available in the 1.4.0

And feel free to  get in touch if you want to contribute more, either with code or any other input ! Thank you again ! 

-Adrian

PR #1215, opened 2022-03-12
HiDPI icon support making sure all icons are scalable

nanoant:
This patch upgrades most of UI parts than are using Qt methods providing HiDPI support to scalable SVG icons. Making the Scopy interface look more polished and sharp on Retina displays.

This patch also does major cleanup in `resources/` subfolder, removing some temporary stray files, consolidating file naming and making sure default (dark) and light theme assets are consistent.

This is achieved by dedicated `resources/generate.py` script that finds all referenced icon assets and generates entries for Qt RCC.

Please click on following image to see it at full scale:
image

Certain UI elements despite using SVG assets are however still not sharp on HiDPI display because Qt does not support HiDPI SVG assets on `border-image` and `background-image`. Only `icon` and `image` are properly scaled. Therefore these parts would need to be rewritten using `image` and `icon`.

image


adisuciu:
This looks great !  We can merge this in the next release candidate.

Few questions:
1. Can you please signoff your commits ?
2. in [1dec7c4](https://github.com/analogdevicesinc/scopy/pull/1215/commits/1dec7c4373857db89e7a2ca7f766653e9c594519) you mention that  

> Remaining (to be fixed) problems:
> 
> 1. setIconSize(32, 32) works differently with scalable graphics, scaling
>    icons to full 32x32 size. Therefore in few places this was replaced
>    by setIconSize(20, 20) for buttons, but icon padding was lost.
> 
> 2. background-image in qss which is used as a workaround to put the icon
>    on left or right of the button is not rendered with HiDPI despite using
>    scalable SVG sources.
>    We should stop using background-image workaround.

Have these been fixed with this commit [6a7a0f7](https://github.com/analogdevicesinc/scopy/pull/1215/commits/6a7a0f7629a37759f5098614096e1887897aec94) and [b6d43c0](https://github.com/analogdevicesinc/scopy/pull/1215/commits/b6d43c0c4ead8287edf4e6c2790b089554487fb5) or ar they still TODO ?

3. Also if I understand correctly, the script scrapes the cpp files and finds references to the svg files. From these references it generates rcc-compatible entries that can be copied and pasted to the resources.qrc. I don't understand what the unneeded part is though - does it try to find svg files that are not referenced anywhere in the source code ?

-Adrian

adisuciu (approved):

adisuciu (approved):

adisuciu:
@nanoant I will merge this sometime this week. Let me know if you have any additional comments.

nanoant:
Hi @adisuciu. Sorry I missed GitHub notifications. Here are my answers to your questions.

> 1. Can you please signoff your commits ?

They all have GPG signature attached for my GH account.

> 2. in 1dec7c4 you mention that (...)

b6d43c0c4ead8287edf4e6c2790b089554487fb5 reintroduced lost padding, but `background-image` problem was not completely resolved. I think proper solution would be to add proper support for `background-image` scalable graphics in Qt, as they did for other QSS tags. And I did not find any easy way to convert it to `image`.

FYI Qt introduced handling for `@2x` for `background-image` but it did not introduce handling for scalable graphics. I had a Qt commit sha for that somewhere.

> 3. Also if I understand correctly, the script scrapes the cpp files and finds references to the svg files. From these references it generates rcc-compatible entries that can be copied and pasted to the resources.qrc. I don't understand what the unneeded part is though - does it try to find svg files that are not referenced anywhere in the source code ?

It will report both unneeded and missing files, otherwise is generates partial XML that can be pasted to .qrc file, so it is useful to keep `resources/` tidy. It uses specific set of regex rules to define how to find the files:

~~~py
        ('**/*.qss;**/*.qss.c', r'(?<=url\(:/).*?(?=\))'),
        ('**/*.html*',          r'(?<=src="qrc:/).*?(?=")'),
        ('**/*.ui',             r'(?<=>:/).*?(?=<)'),
        ('**/*.ui',             r'(?<=url\(:/).*?(?=\))'),
        ('**/*.cpp;**/*.cc',    r'(?<=":/).*?\.(?:svg|png)(?=")'),
        ('**/*.cpp;**/*.cc',    r'(?<=QIcon::fromTheme\(").*?(?=");icons/scopy-default/icons/%s.svg;icons/scopy-light/icons/%s.svg'),
        ('src/toolmenu.cpp',    r'(?<=<< ").*(?=");icons/scopy-default/icons/%s.svg;icons/scopy-light/icons/%s.svg'),
~~~

I think & hope I covered all of the cases there.

adisuciu:
Awesome thanks. So the only TODO moving forward would be to find a way to not use background image in qss and use image so things are scaled nicely. Right ? I'll create an issue for that if that's the case.

nanoant:
> So the only TODO moving forward would be to find a way to not use background image in qss and use image so things are scaled nicely. Right ? I'll create an issue for that if that's the case.

That's right.

adisuciu:
Created https://github.com/analogdevicesinc/scopy/issues/1317

adisuciu:
@nanoant  is it possible to include the CMakeLists.txt and AndroidManifest.xml in the scrape ? The icon files are used to generate .desktop files for linux so the application actually has icons.

nanoant:
> @nanoant is it possible to include the CMakeLists.txt and AndroidManifest.xml in the scrape ? The icon files are used to generate .desktop files for linux so the application actually has icons.

Sure. I managed to restore `icon_small.svg` for `share/icons/hicolor/apps/scalable/scopy.svg` but all Android native files are in `android/res/` which I did not touch, since the script is only intended for `resources/`. I've sent PR https://github.com/analogdevicesinc/scopy/pull/1319.

adisuciu:
Yes, you are correct about the Android files. Thanks for the quick reply !

Issue #18, opened 2016-10-24
Oscilloscope: User selectable waveform and background colors


dNechita:
Moving this from a list of issues at #9 into its own issue.

Initial issue:
"Are there plans to have the waveform color changeable by the user? Also have the background be able to be black or white? I like the black background personally but for screen shots that you might want to include in a report document ( that gets printed on paper ) you might want a white background."

dNechita:
Need mokcup for the color chooser widget.

pcercuei:
About the background color - maybe that's something the theme should provide?
Then we can have a dark theme (the current one) and a light theme.

dNechita:
Yes, the background color should be provided by the theme.
In case we need to export a .png of the plot, we should add options like "Black and white" or "Invert colors" that would make the background white in case it is needed for printing.

rgetz:
+1 for comment about printing with white background. I think we are running into this as people make up their labs.

-Robin

rgetz:
for managing preferences, I think you just have a list of features and a choose able colour by each. Something similar to (but not exactly); I still think we should keep a "screen" theme, and a "print" theme (which might be dark/light by default).
![image](https://user-images.githubusercontent.com/1017869/96823804-4e852600-13fb-11eb-9bf4-b895eff65407.png)


for picking colours, I think we would just use [QColorDialog](https://doc.qt.io/qt-5/qcolordialog.html) (or similar)
![image](https://user-images.githubusercontent.com/1017869/96823539-b4bd7900-13fa-11eb-8df1-cb259d8d5891.png)

adisuciu:
We could do that, we just need to figure out what our features are. 

Another quick and dirty way to do it is with stylesheet files. Most of the stylesheets are part of a single file(global.qss) already. We could load similar files at runtime and have the file bundled with the executable as failsafe. This way you can create your own themes - or have a selection inbetween multiple stylesheet files that are provided with Scopy - similar to what we did for languages. It's less user friendly but more customizable.

rgetz:
it's a tradeoff between making it customizable, vs how difficult it is to customize...

I don't want to be in the business of making customizations for everyone, because it's too hard for users. (as I'm sure you don't either)...

connorka:
In addition to printing, slides are best done with light colors, which is also generally true for videos. I agree with Robin that one does not need a lot of options, at least for backgrounds. For signals/line plots etc, having the ability to choose colors, as one does with Excel and Matlab, can be very useful.

[...]

PR #1315, opened 2022-04-05
Support for changing logic analzyer trace colors

frmdstryr:
This adds a combobox to the logic analyzer channel settings with Qt's [SVG color names](https://doc.qt.io/qt-5/qcolorconstants.html#svg-colors) . It would be nice if it can be hooked up to the preferences but I haven't looked into how to do that. 

image

Somewhat related to https://github.com/analogdevicesinc/scopy/issues/18.

I'm not sure how to test the mixed signal view, it does not enable for me.  

Feedback is welcome.


adisuciu:
That's awesome ! Thanks ! I'll give it a test in the following days and get back at you ..

adisuciu:
This is very nice and clean ! I'll look into adding the same thing to the oscilloscope. Before merging, can you amend your commit with a --signoff tag ? 
Thanks !
-Adrian

frmdstryr:
Never had to use signoff before so hopefully that was correct.

adisuciu:
You did it well the first time here:
[](https://github.com/frmdstryr)@[frmdstryr](https://github.com/frmdstryr) frmdstryr [force-pushed ](https://github.com/analogdevicesinc/scopy/compare/3abf38b19e13f9be7f10efad0b5534c1aa597e03..ea2b76da3f83440aa5f224ae82d4ef53f8ddc59f)the logic-analyzer-trace-colors branch from [3abf38b ](https://github.com/analogdevicesinc/scopy/commit/3abf38b19e13f9be7f10efad0b5534c1aa597e03)to [ea2b76d ](https://github.com/analogdevicesinc/scopy/commit/ea2b76da3f83440aa5f224ae82d4ef53f8ddc59f)[27 minutes ago](https://github.com/analogdevicesinc/scopy/pull/1315#event-6387637595)

frmdstryr:
Alright changed it back.

It might also be a good idea to make the trace handle colors to match the line color. Should I add that to this or do another PR or would you want to do that?

adisuciu:
Having selective waveforms have a different color make sense if you want to highlight them. Changing the color of the handle is more involved because of a few reasons:

The cables that are bundled with the M2k should match the handle color in the software. This is what's currently implemented.
Synchronizing the handle color to the trace color would would mean that when initialized all the handles are either all blue which makes it confusing I think. The other option would be to match the waveform colors to the handles which would mean that each line would have a different color, making it a little distracting. We'll talk about this internally.

frmdstryr:
At 16 channels everything is distracting..  It'd be nice if a tooltip with the trace name would show up when hovering over a trace. 

If you look at different MSO pictures a it seems most default all the LA channels to the same color. I think the handle should match the trace at least.  Maybe it could be a setting in the preferences?

adisuciu:
Yes, I agree that everything is distracting at 16 channels :)  .. It should be a setting. 

Current behavior (handle colors fixed,trace colors settable)  
vs
match handles and trace colors. When activated you get the multicolored traces from the first picture I submitted. I think that would be best. If you need help adding the preference in the preference menu, I can help with that. 
I will merge this PR and you can submit a new one if you want.

Thanks for your help !

adisuciu (approved):

frmdstryr:
Thank you!

PR #1370, opened 2022-09-02
Implemented decoder table for logic analyzer

andrei47w:
Groups decoders by messages
- cannot run LA when table is open, only single button is available
- information about decoder messages (sample, duration) can be enabled in preferences
- clicking rows in table will zoom in the plot on the message
- clicking annotations in plot will select the equivalent row in table (if available)

settings:
- selectable decoder
- selectable primary annotation type (groups annotations by the specified type)
- filter unwanted annotations
- regex search (only searches through what is visible with the current settings)

work in progress...

https://user-images.githubusercontent.com/92979643/188131356-e7c6426e-4121-40af-b8aa-800db21ee86e.mp4

### DIFF
--- a/src/logicanalyzer/annotationcurve.cpp
+++ b/src/logicanalyzer/annotationcurve.cpp
@@ -544,7 +544,7 @@ void AnnotationCurve::drawAnnotationInfo(int row, uint64_t start, uint64_t end, 
 	if (start != end) {
 		rect = QRectF(QPointF(xMap.transform(fromSampleToTime(start)), yMap.transform(offset)), QSizeF(width, m_infoHeight));
 	} else {
-		rect = QRectF(QPointF(canvasRect.x() + canvasRect.width() * infoWidthRatio, yMap.transform(offset)), QSizeF(canvasRect.width() * (1 - infoWidthRatio), m_infoHeight));
+		rect = QRectF(QPointF(canvasRect.x() + canvasRect.width() * infoWidthRatio, yMap.transform(offset)), QSizeF(canvasRect.width() * (1 - infoWidthRatio * 2), m_infoHeight));
 	}
 
 	painter->save();
--- a/src/logicanalyzer/annotationcurve.cpp
+++ b/src/logicanalyzer/annotationcurve.cpp
@@ -509,6 +509,69 @@ void AnnotationCurve::drawBlock(int row, uint64_t start, uint64_t end, QPainter 
     painter->restore();
 }
 
+QString AnnotationCurve::formatSeconds(double sec) const
+{
+	QString string;
+	auto time = abs(sec);
+	if (time > 1) {
+		string = QString("%1 s").arg(sec);
+	}
+	else if (time > 0.001) {
+		string = QString("%1 ms").arg(sec * 10E+2);
+	}
+	else if (time > 0.000001) {
+		string = QString("%1 μs").arg(sec * 10E+5);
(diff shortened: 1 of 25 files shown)


rgetz:
This does indeed look like a super useful feature - however - there are alot of commits in there, that look like they could be in a separate clean up PR (all the initial remove warnings, and remove unused vars?) and the later "bug fix" ones should be squashed into the initial code that was changed with the bug?

Thanks
-Robin

frmdstryr:
I just gave it a quick test. Overall it looks looks great! The grouping works as expected, as does the filtering of messages. The search is really nice, that will really come in handy.

One area that could be improved is the loading speed. Could the processing be pushed into a thread or done in chunks like the decoding appears to be? At a 100Msps sample rate, a 10M sample usb capture takes about 30 seconds to open the panel.  I usually do 100M captures to test a device enumeration.  If it's linear time that would be almost 5 minutes...

All in all nice work!

andrei47w:
I added a commit which improves table loading times and I will work on organizing the commits. Thank you for the feedback.

-Andrei

src/logicanalyzer/annotationcurve.cpp
```diff
@@ -500,20 +512,107 @@ void AnnotationCurve::drawBlock(int row, uint64_t start, uint64_t end, QPainter
...
+					 const QRectF &canvasRect) const {
+
+	const double infoHeightInPoints = yMap.invTransform(m_infoHeight) - yMap.invTransform(0);
+	const double infoWidthRatio = 0.09;
+	double HeightInPoints = yMap.invTransform(m_traceHeight) - yMap.invTransform(0);
+	double offset = m_pixelOffset + (row) * (HeightInPoints) - infoHeightInPoints;
+
+	double width = xMap.transform(fromSampleToTime(end)) - xMap.transform(fromSampleToTime(start));
+
+	const int r = 20 / 4;
```
AlexandraTrifan, line 546: Do these values have an explanation?

src/logicanalyzer/logic_analyzer.cpp
```diff
@@ -1451,12 +1512,237 @@ void LogicAnalyzer::setupUi()
 	m_exportSettings->disableUIMargins();
 	connect(m_exportSettings->getExportButton(), &QPushButton::clicked,
 		this, &LogicAnalyzer::exportData);
+
+
+	// Filter in decoder table
+	filterMessages = new DropdownSwitchList();
```
AlexandraTrifan, line 1518: should this have the LA as a parent?

src/logicanalyzer/logic_analyzer.cpp
```diff
@@ -2201,6 +2539,26 @@ void LogicAnalyzer::setupDecoders()
...
+			std::string title = result.ann->row()->title().toStdString();
+			title = title.erase(0, title.find(':') + 1);
+
+			int row_index = ui->PrimaryAnnotationComboBox->findText(QString::fromStdString(title));
+			if (row_index != -1) {
+				int row = 0;
+				if (getGroupOffset() < result.index) {
+					row = result.index / getGroupSize();
+				}
+				ui->PrimaryAnnotationComboBox->setCurrentIndex(row_index);
```
AlexandraTrifan, line 2558: not functional: for consistency, this could be lowercase

AlexandraTrifan (approved):

PR #1401, opened 2023-06-13
Scopy2 new add page

andreidanila1:


core/src/scopyhomeaddpage.cpp
```diff
@@ -1,68 +1,306 @@
...
+	QString uri = ui->editUri->text();
+	deviceImpl = DeviceFactory::build(uri, pluginManager, "");
+	DeviceLoader* dl = new DeviceLoader(deviceImpl, this);
+	dl->init();
+
+	connect(dl, &DeviceLoader::initialized, this, &ScopyHomeAddPage::deviceLoaderInitialized);
+	connect(dl, &DeviceLoader::initialized, dl, &QObject::deleteLater); // don't forget to delete loader once we're done
+}
+
+void ScopyHomeAddPage::removePluginsCheckBoxs()
```
AlexandraTrifan, line 286: typo

core/src/scopyhomeaddpage.cpp
```diff
@@ -1,68 +1,306 @@
...
+{
+	if (ui->subSeparatorPlugins->getContentWidget()->layout()) {
+		QLayoutItem* item;
+		while ((item = ui->subSeparatorPlugins->getContentWidget()->layout()->takeAt(0)) != nullptr) {
```
AlexandraTrifan, line 290: will this throw index out of range at some point?

core/src/scopyhomeaddpage.cpp
```diff
@@ -1,68 +1,306 @@
...
+		QHBoxLayout *hLayout = new QHBoxLayout(pluginInfo);
+
+		QCheckBox *cb = new QCheckBox();
+		cb->setText(p->name());
+		cb->setChecked(p->enabled());
+		connect(cb, &QCheckBox::toggled, this, [=](bool en){
+			p->setEnabled(en);
+		});
+
+		QLabel *descriptionLabel = new QLabel();
```
AlexandraTrifan, line 260: same as above

core/src/scopyhomeaddpage.cpp
```diff
@@ -1,68 +1,306 @@
...
+			QCheckBox *cb = new QCheckBox();
+			cb->setText(backEnd);
+			if (backEnd.compare("serial") == 0) {
+				connect(cb, &QCheckBox::toggled, this, [=](bool en) {
+					if (en) {
+						scanParamsList.push_back(backEnd + ":");
+					} else {
+						scanParamsList.removeOne(backEnd + ":");
+					}
+					ui->serialSettingsWidget->setEnabled(en);
```
adisuciu, line 163: You could create the connect when you create the checkbox instead of dynamically connecting here.
Here you can just enable/disable the control

core/src/scopyhomeaddpage.cpp
```diff
@@ -1,68 +1,306 @@
...
+	}
 }
 
+void ScopyHomeAddPage::verifyIioBackend()
+{
+	bool scan = false;
+	int backEndsCount = iio_get_backends_count();
+	for (int i = 0; i < backEndsCount; i++) {
+		QString backEnd(iio_get_backend(i));
+		if (backEnd.compare("xml")) {
```
adisuciu, line 153: Why are you checking for xml backend ?

core/src/scopyhomeaddpage.cpp
```diff
@@ -1,68 +1,306 @@
...
+					}
+					ui->serialSettingsWidget->setEnabled(en);
+				});
+			} else {
+				connect(cb, &QCheckBox::toggled, this, [=](bool en) {
+					if (en) {
+						scanParamsList.push_back(backEnd + ":");
+					} else {
+						scanParamsList.removeOne(backEnd + ":");
+					}
```
adisuciu, line 171: This connect does the same thing in both cases

core/src/scopyhomeaddpage.cpp
```diff
@@ -1,68 +1,306 @@
...
+	}
+	Q_EMIT contextDataAvailable(contextInfo);
+}
 
 bool ScopyHomeAddPage::verify() {
 	QString uri = ui->editUri->text();
 	struct iio_context* ctx = ContextProvider::GetInstance()->open(uri);
-	if(ctx) {		
+	if(ctx) {
+		getContextData(ctx);
```
adisuciu, line 200: emit the result from here instead of relying on the getContextData() to emit

adisuciu (changes requested):

core/src/scopyhomeaddpage.cpp
```diff
@@ -1,68 +1,306 @@
...
+			deviceInfoPage->update(key, ctxInfo[key]);
+		}
+	},Qt::QueuedConnection);
+}
+
+ScopyHomeAddPage::~ScopyHomeAddPage()
+{
+	delete ui;
+	if (deviceImpl) {
+		delete deviceImpl;
```
AlexandraTrifan, line 90: crashes here, needs some investigation

core/src/scopyhomeaddpage.cpp
andreidanila1: I thought it should be treated differently from the others

core/src/scopyhomeaddpage.cpp
andreidanila1: It should do this when you delete all the items from a layout.

PR #1413, opened 2023-07-04
Scopy2 regmap 

IonutMuthi:
register map functionality


plugins/regmap/src/deviceregistermap.cpp
```diff
@@ -2,115 +2,168 @@
...
-        layout->addWidget(searchBarWidget);
-        //		layout->addWidget(registerMapTableWidget);
-        layout->addWidget(recycerViewAdapter->getWidget());
-
+        registerMapTableLayout->addWidget(tableHeadWidget);
+        auto aux = registerMapTableWidget->getWidget();
+        if (aux) {
+            registerMapTableLayout->addWidget(aux);
+        }
+        layout->addWidget(registerMapTable,5);
```
AlexandraTrifan, line 73: what's the meaning of 5?

PR #1431, opened 2023-07-26
Dev emu add page

andreidanila1:


core/src/emuwidget.cpp
```diff
@@ -0,0 +1,187 @@
...
+		arguments.append(m_ui->lineEditXmlPath->text());
+		arguments.append(m_ui->lineEditRxTxDevice->text());
+	}
+	return arguments;
+}
+
+QString EmuWidget::findEmuPath()
+{
+	Preferences *p = Preferences::GetInstance();
+	QString pathFromPreferinces = p->get("iio_emu_path").toString();
```
adisuciu, line 108: Is this not initialized ?

adisuciu, line 108: typo

core/src/emuwidget.cpp
```diff
@@ -0,0 +1,187 @@
...
+	pathFromPreferinces.replace(" ", "");
```
adisuciu, line 109: why replace ?

core/src/emuwidget.cpp
```diff
@@ -0,0 +1,187 @@
...
+	// iio-emu found in system
+	QString program = qgetenv("IIOEMU_BIN");
+	if (program.isEmpty()) {
+		// Env Var not found, try local folder
+		QString dirPath = QCoreApplication::applicationDirPath();
+		program = dirPath + "/iio-emu";
+#if defined(_WIN32) || defined(__CYGWIN__) || defined(__MINGW32__)
+		program += ".exe";
```
adisuciu, line 117: Did you test this on windows ?

core/src/emuwidget.cpp
```diff
@@ -0,0 +1,187 @@
...
+QString EmuWidget::findEmuPath()
+{
+	Preferences *p = Preferences::GetInstance();
+	QString pathFromPreferinces = p->get("iio_emu_path").toString();
+	pathFromPreferinces.replace(" ", "");
+	// iio-emu found in system
+	QString program = qgetenv("IIOEMU_BIN");
+	if (program.isEmpty()) {
+		// Env Var not found, try local folder
+		QString dirPath = QCoreApplication::applicationDirPath();
```
adisuciu, line 114: this flow does not account for iio-emu being part of PATH - you just run it as

iio-emu

core/src/scopyhomeaddpage.cpp
```diff
@@ -94,21 +104,26 @@ void ScopyHomeAddPage::initAddPage()
 	scanIcon->setFileName(":/gui/loading.gif");
 	ui->btnScan->setAnimation(scanIcon);
 
-	QRegExp re("[5-9]{1}(n|o|e|m|s){1}[1-2]{1}(x|r|d){1}$");
+	QRegExp re("[5-9]{1}(n|o|e|m|s){1}[1-2]{1}(x|r|d){0,1}$");
```
adisuciu, line 107: Why do we need this ?

core/src/scopyhomepage.cpp
```diff
@@ -71,6 +72,16 @@ void ScopyHomePage::disconnectDevice(QString id) {
 	db->disconnectDevice(id);
 }
 
+void ScopyHomePage::disconnectEmuDevice(QString id)
```
adisuciu, line 75: I don't think we should do this. We use the emulator for development. We can use it to simulate loss of connection with the board by disabling the emulator. 
Therefore the plugins should gracefully handle this, and not be handled by the home page (by forcefully disconnecting the device)

core/src/scopyhomeaddpage.cpp
```diff
@@ -275,12 +290,24 @@ void ScopyHomeAddPage::deviceLoaderInitialized()
 	}
 	ui->stackedWidget->setCurrentWidget(ui->deviceInfoPage);
 	ui->btnVerify->stopAnimation();
+	if (emuWidget->getEmuDeviceUri().compare(deviceImpl->param()) == 0) {
```
adisuciu, line 293: Is this needed ?

adisuciu (changes requested):

PR #1438, opened 2023-08-08
Loading preferences on restart bug fix

andrei47w:
- preferences are now saved on aboutToQuit signal instead of inside the destructor
- made ApplicationRestarter a singleton

### DIFF
--- a/core/src/application_restarter.cpp
+++ b/core/src/application_restarter.cpp
@@ -10,11 +10,19 @@
 
 using namespace scopy;
 
-ApplicationRestarter::ApplicationRestarter(const QString &executable)
-        : m_executable(executable)
-        , m_currentPath(QDir::currentPath())
-{
+ApplicationRestarter* ApplicationRestarter::pinstance_{nullptr};
 
+ApplicationRestarter::ApplicationRestarter(const QString &executable)
+	: m_executable(executable)
+	, m_currentPath(QDir::currentPath())
+	, m_restart(false)
+{
+	pinstance_ = this;
+}
+
+ApplicationRestarter *ApplicationRestarter::GetInstance()
+{
+	return pinstance_;
 }
 
 void ApplicationRestarter::setArguments(const QStringList &arguments)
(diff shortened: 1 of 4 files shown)


tmp/src/scopy_color_editor.cpp
```diff
@@ -110,7 +110,7 @@ ScopyColorEditor::ScopyColorEditor(QApplication *app, QWidget *parent)
 		m_ui->saveBtn->setEnabled(true);
 	});
 	connect(m_ui->restartBtn, &QPushButton::clicked, [=](){
-		adiscope::ApplicationRestarter::triggerRestart();
+		adiscope::ApplicationRestarter::GetInstance()->triggerRestart();
```
adisuciu, line 113: This is not compiled, it will eventually be removed from the project. No need to update here ..

pluginbase/include/pluginbase/preferences.h
```diff
@@ -65,6 +60,13 @@ class SCOPY_PLUGINBASE_EXPORT Preferences : public QObject
 	 */
 	void setPreferencesFilename(QString s);
 
+public Q_SLOTS:
```
adisuciu, line 63: You can make both save and load as QSLOTS

core/src/application_restarter.cpp
```diff
@@ -43,6 +51,6 @@ int ApplicationRestarter::restart(int exitCode)
 
 void ApplicationRestarter::triggerRestart()
 {
-	qApp->setProperty("restart", QVariant(true));
+	m_restart = true;
```
adisuciu, line 54: You can make this method static by getting the GetInstance() inside the method. Makes it easier to call ..

core/src/scopypreferencespage.cpp
```diff
@@ -57,7 +57,7 @@ QWidget* ScopyPreferencesPage::createRestartWidget() {
 	lay->addSpacerItem(space);
 	lay->addWidget(btn,1);
 
-	w->connect(btn,&QPushButton::clicked, btn, [](){ApplicationRestarter::triggerRestart();});
+	w->connect(btn,&QPushButton::clicked, btn, [](){ApplicationRestarter::GetInstance()->triggerRestart();});
```
adisuciu, line 60: By making the triggerRestart method static it makes it easier to call and improves readability.

adisuciu (changes requested):

PR #1445, opened 2023-08-18
About page fixes

andrei47w:


core/src/scopyaboutpage.cpp
```diff
@@ -17,24 +17,64 @@ QWidget* ScopyAboutPage::buildPage(QString src, QTextDocument::ResourceType r) {
...
-	case QTextDocument::HtmlResource:
-	default:
-		browser->setSource(src,r);
-		break;
-	}
-	QPushButton *backButton = new QPushButton("Back",page);
-	backButton->setProperty("blue_button",true);
-	lay->addWidget(backButton);
-	connect(backButton,SIGNAL(clicked()),browser,SLOT(backward()));
+	browser->setSource(src,r);
```
adisuciu, line 28: Please review this
This implementation requires that the plugin developer creates an explicit "about" file, and cannot rely on simply returning a string.

adisuciu (changes requested):

adisuciu (approved):

PR #1468, opened 2023-09-14
Generic popup

Andrei-Fabian-Pop:


gui/src/widgets/popupwidget.cpp
```diff
@@ -0,0 +1,95 @@
...
+	ui(new Ui::PopupWidget),
+	QWidget(parent)
+{
+	ui->setupUi(this);
+
+	StyleHelper::initColorMap();
+	StyleHelper::BlueButton(ui->btnContinue, "btnContinue");
+	StyleHelper::BlueButton(ui->btnExit, "btnExit");
+	StyleHelper::OverlayMenu(this, "popupOverlay");
+	ui->title->setStyleSheet("background-color: #101014; border: 1px solid #070708"); // color match the description QTextBrowser
```
adisuciu, line 18: This code is duplicated. Why do you need 2 constructors ? Can you just create a constructor and a setDescription method ?

adisuciu, line 18: This belongs to a stylehelper. Create a stylehelper with an appropriate name.

gui/src/widgets/popupwidget.cpp
```diff
@@ -0,0 +1,95 @@
...
+
+using namespace scopy::gui;
+
+PopupWidget::PopupWidget(QWidget* parent) :
+	ui(new Ui::PopupWidget),
+	QWidget(parent)
+{
+	ui->setupUi(this);
+
+	StyleHelper::initColorMap();
```
adisuciu, line 14: The stylehelper initColorMap should not be called for every popup widget that is constructed.

plugins/swiot/src/swiotplugin.cpp
```diff
@@ -343,6 +343,7 @@ void SWIOTPlugin::startTutorial() {
 		auto configTool = configTme->tool();
 		auto parent = Util::findContainingWindow(configTool);
 		auto tut = new gui::TutorialBuilder(configTool, ":/swiot/tutorial_chapters.json", "config", parent);
+		tut->setTitle("CONFIG ");
```
adisuciu, line 346: Do you need the space in "CONFIG " ? If yes, it could suggest that a style change is needed to acomodate spacing.

gui/src/tutorialoverlay.cpp
```diff
@@ -181,21 +175,13 @@ void TutorialOverlay::buildUi()
...
-	ui->description->setStyleSheet("background-color: rgba(0,0,0,60);color: white;");
-	ui->title->setStyleSheet("color:white");
-	ui->btnExit->setStyleSheet("width:80;height:20");
-	ui->btnExit->setProperty("blue_button", true);
-	ui->btnNext->setStyleSheet("width:80;height:20");
-	ui->btnNext->setProperty("blue_button", true);
-	ui->btnNext->setFocus();
-	connect(ui->btnNext,&QPushButton::clicked,this,&TutorialOverlay::next);
-	connect(ui->btnExit, &QPushButton::clicked, this, &TutorialOverlay::abort);
+	m_popupWidget = new PopupWidget();
```
adisuciu, line 178: Are you reusing the same popup widget for each of the tutorial steps ? 
Wouldn't it be better if each step had it's own popup ? I think it already has it's own hoverwidget ..

adisuciu (changes requested):

gui/src/widgets/popupwidget.cpp
Andrei-Fabian-Pop: Should the color codes also have an entry in the StyleHelper colorMap, or is it safe to assume that because the codes are not used anywhere else the string can be left "as is"?

adisuciu: Yes they should be a part of the colormap with apropriate naming.

adisuciu (approved):

PR #1481, opened 2023-09-25
Core: Class diagram updates

andreidanila1:


adisuciu:
Can you squash these to a single commit ?

PR #1498, opened 2023-10-19
Dev status bar

Andrei-Fabian-Pop:


gui/src/widgets/scopystatusbar.cpp
```diff
@@ -0,0 +1,77 @@
...
+Q_LOGGING_CATEGORY(CAT_SCOPYSTATUSBAR, "ScopyStatusBar")
+
+ScopyStatusBar::ScopyStatusBar(QWidget *parent)
+	: MenuVAnim(parent)
+{
+	initUi();
+
+	auto statusManager = StatusManager::GetInstance();
+	connect(statusManager, &StatusManager::sendStatus, this,
+		[this](StatusMessage *message) { displayStatus(message); });
```
adisuciu, line 20: Does this need to be a lambda ?

gui/src/widgets/scopystatusbar.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+	connect(statusManager, &StatusManager::clearDisplay, this, [this]() { clearStatus(); });
```
adisuciu, line 22: Does this need to be a lambda ?

gui/include/gui/widgets/scopystatusbar.h
```diff
@@ -0,0 +1,45 @@
...
+	Q_OBJECT
+	QWIDGET_PAINT_EVENT_HELPER
+public:
+	explicit ScopyStatusBar(QWidget *parent = nullptr);
+
+Q_SIGNALS:
+	void requestHistory();
+
+public Q_SLOTS:
+	void displayStatus(StatusMessage *statusMessage);
```
adisuciu, line 28: Rename this to displayStatusMessage, and clearStatusMessage

pluginbase/include/pluginbase/statusmanager.h
```diff
@@ -0,0 +1,73 @@
...
+
+#define NUMBER_OF_RETAINED_MESSAGES 10
+#define DEFAULT_DISPLAY_TIME 5000
+
+namespace scopy {
+class SCOPY_PLUGINBASE_EXPORT StatusManager : public QObject
+{
+	Q_OBJECT
+protected:
+	StatusManager(QObject *parent = nullptr);
```
adisuciu, line 19: rename to StatusBarManager ?

pluginbase/include/pluginbase/statusmanager.h
```diff
@@ -0,0 +1,73 @@
...
+	StatusManager(StatusManager &other) = delete;
+	void operator=(const StatusManager &) = delete;
+	static StatusManager *GetInstance();
+
+	/**
+	 * @brief Ads temporary message in the queue to be displayed, when possible, in the Scopy status bar.
+	 * @param message QString with the message to be displayed
+	 * @param ms The time that the message will be displayed (in milliseconds)
+	 * */
+	void addTemporaryMessage(const QString &message, int ms = DEFAULT_DISPLAY_TIME);
```
adisuciu, line 33: pushMessage

pluginbase/include/pluginbase/statusmanager.h
```diff
@@ -0,0 +1,73 @@
...
+	 * @param ms The time that the widget will be displayed (in milliseconds)
+	 * */
+	void addTemporaryWidget(QWidget *widget, QString title, int ms = -1);
+
+	/**
+	 * @brief Overrides any message currently displayed with the message sent as parameter
+	 * @param message QString with the urgent message
+	 * @param ms The time that the message will be displayed (in milliseconds)
+	 * */
+	void addUrgentMessage(const QString &message, int ms = DEFAULT_DISPLAY_TIME);
```
adisuciu, line 50: pushUrgentMessage

pluginbase/src/statusmessage.cpp
```diff
@@ -0,0 +1,57 @@
...
+}
+
+QString StatusMessage::getText() { return m_text; }
+
+void StatusMessage::setText(QString text) { m_text = std::move(text); }
+
+QWidget *StatusMessage::getWidget()
+{
+	if(!m_widget) {
+		auto label = new QLabel(this);
```
adisuciu, line 31: Do you need to do this on every get ? Why not just do it in the constructor ?

pluginbase/src/statusmessage.cpp
```diff
@@ -0,0 +1,57 @@
...
+{
+	if(!m_permanent) {
+		delete m_widget;
+		m_widget = nullptr;
+	}
+}
+
+QString StatusMessage::getText() { return m_text; }
+
+void StatusMessage::setText(QString text) { m_text = std::move(text); }
```
adisuciu, line 26: Huh ?

pluginbase/src/statusmessage.cpp
```diff
@@ -0,0 +1,57 @@
...
+		label->setSizePolicy(QSizePolicy::Expanding, QSizePolicy::Fixed);
+		label->setMaximumHeight(15);
+
+		m_widget = label;
+	}
+
+	return m_widget;
+}
+
+void StatusMessage::setWidget(QWidget *widget) { m_widget = widget; }
```
adisuciu, line 44: These should be handled in the constructor ? .. not changeable after construction ?

pluginbase/src/statusmessage.cpp
```diff
@@ -0,0 +1,57 @@
...
+
+int StatusMessage::getDisplayTime() { return m_ms; }
+
+void StatusMessage::setDisplayTime(int ms) { m_ms = ms; }
```
adisuciu, line 48: These should be handled in the constructor ? .. not changeable after construction ?

pluginbase/src/statusmanager.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+StatusManager *StatusManager::GetInstance()
+{
+	if(pinstance_ == nullptr) {
+		pinstance_ = new StatusManager(QApplication::instance()); // singleton has the app as parent
+	}
+	return pinstance_;
+}
+
+void StatusManager::addTemporaryMessage(const QString &message, int ms)
```
adisuciu, line 35: To simply call StatusBarManager::pushMessage without getting the instance first, you can declare a static pushMessage and a private _pushMessage() member

The static method gets the instance and calls the member. This allows for cleaner singleton interfacing.
Smth similar is implemented here:

https://github.com/analogdevicesinc/scopy/pull/1500/files#diff-00efcbdc18feab83b204658845a8715cbf6d323ffd89c7216d3a866e2ff73ca2R31-R34

adisuciu (changes requested):

pluginbase/src/statusmessage.cpp
Andrei-Fabian-Pop: The status message can contain either a QString or a QWidget*, if we want to have the widget unchangeable we could make 2 constructors with the difference that one takes only the QString and one takes only the QWidget*. How does this sound?

PR #1519, opened 2023-12-20
IIO Widgets

Andrei-Fabian-Pop:
Added infrastructure for generating IIO Widgets that tie to an iio structure and displays its properties


iio-widgets/src/datastrategy/channelattrdatastrategy.cpp
```diff
@@ -0,0 +1,61 @@
+#include "datastrategy/channelattrdatastrategy.h"
+#include <QFile>
+
+#define BUFFER_SIZE 256
```
adisuciu, line 4: Buffer size attribute should be a lot bigger than 256 - we could try 16k. Apply this to all iio string buffer sizes

iio-widgets/src/datastrategy/channelattrdatastrategy.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+Q_LOGGING_CATEGORY(CAT_ATTR_DATA_STRATEGY, "AttrDataStrategy")
+
+using namespace scopy::attr;
```
adisuciu, line 8: Is there a need to add an extra namespace ? I would just keep it under 
namespace scopy { }

iio-widgets/src/datastrategy/filedemodatastrategy.cpp
```diff
@@ -0,0 +1,27 @@
+#include "datastrategy/filedemodatastrategy.h"
+#include <utility>
+#include <QFile>
+
+using namespace scopy::attr;
+
+FileDemoDataStrategy::FileDemoDataStrategy(IIOWidgetFactoryRecipe recipe, QObject *parent)
+{
+	m_recipe = std::move(recipe);
```
adisuciu, line 9: std::move caused issues previously. Are you sure this is needed here ? Make sure you update this to all necessary classes.

iio-widgets/src/datastrategy/deviceattrdatastrategy.cpp
```diff
@@ -0,0 +1,57 @@
...
+#define BUFFER_SIZE 256
+
+Q_LOGGING_CATEGORY(CAT_DEVICE_DATA_STRATEGY, "DeviceDataStrategy")
+
+using namespace scopy::attr;
+
+DeviceAttrDataStrategy::DeviceAttrDataStrategy(IIOWidgetFactoryRecipe recipe, QObject *parent)
+{
+	m_recipe = recipe;
+	setParent(parent);
```
adisuciu, line 11: A better approach here would be to create the underlying QObject before the constructor like this:

DeviceAttrDataStrategy::DeviceAttrDataStrategy(IIOWidgetFactoryRecipe recipe, QObject *parent) : QObject(parent)

Please apply this to all QObjects

plugins/testplugin/src/testplugin.cpp
```diff
@@ -332,7 +332,7 @@ void TestPlugin::initMetadata()
 	{
 	   "priority":2,
 	   "category":[
-	      "test"
+	      "test", "iio"
```
adisuciu, line 335: Test plugin shouldn't have iio category in "production". Using it locally is totally fine.

adisuciu (changes requested):

adisuciu (approved):

PR #1585, opened 2024-03-22
Max refactoring.

andreidanila1:
I decided to make the PR to dev_swiot_refactoring branch in order not to add the new swiot plugin in dev.
This PR is made for checking and testing the functionalities of the max tool (Only max files are relevant).


plugins/swiotrefactor/src/readerthread.cpp
```diff
@@ -0,0 +1,378 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see <https://www.gnu.org/licenses/>.
+ */
+
+#include "readerthread.h"
+
+//#include "src/runtime/ad74413r/bufferlogic.h"
```
Andrei-Fabian-Pop, line 23: i think you can delete this

plugins/swiotrefactor/src/readerthread.cpp
```diff
@@ -0,0 +1,378 @@
...
+			}
+		}
+	} catch(...) {
+		qCritical(CAT_SWIOT_MAX14906) << "Failed to acquire data on DioReaderThread";
+	}
+}
+
+void ReaderThread::addBufferedDevice(iio_device *device) { m_iioDev = device; }
+
+void ReaderThread::initIioChannels()
```
Andrei-Fabian-Pop, line 112: are the commented methods from this class still necessary?

Andrei-Fabian-Pop (changes requested):

plugins/swiotrefactor/CMakeLists.txt
```diff
@@ -0,0 +1,103 @@
...
+elseif(${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
+	set(CMAKE_LIBRARY_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/Scopy.app/Contents/MacOS/plugins/plugins")
+else()
+	set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${SCOPY_PLUGIN_BUILD_PATH})
+endif()
+
+if(ANDROID)
+	find_library(QWT_LIBRARIES REQUIRED NAMES qwt_${ANDROID_ABI})
+else()
+	find_library(QWT_LIBRARIES REQUIRED NAMES qwt)
```
adisuciu, line 72: Is this still required ? Maybe update according to latest plugin template ?

plugins/swiotrefactor/include/swiotrefactor/readerthread.h
```diff
@@ -0,0 +1,111 @@
...
+	//	void onChnlsChange(QMap<int, ChnlInfo *> chnlsInfo);
+	void onSamplingFrequencyComputed(double samplingFreq);
+
+Q_SIGNALS:
+	void readerThreadFinished();
+	void bufferRefilled(QVector<QVector<double>> bufferData, int bufferCounter);
+	void channelDataChanged(int channelId, double value);
+
+private Q_SLOTS:
+	void bufferRefillCommandFinished(scopy::Command *cmd);
```
adisuciu, line 82: scopy:: can be avoided by specifying using namespace scopy within the namespace scopy::swiotrefactor.

That way the scopy namespace is also recognized in the scope of scopy::swiotrefactor

plugins/swiotrefactor/include/swiotrefactor/max14906/diodigitalchannelcontroller.h
```diff
@@ -0,0 +1,75 @@
...
+	void writeCurrentLimitCmdFinished(scopy::Command *cmd);
+	void writeTypeCmdFinished(scopy::Command *cmd);
+
+private:
+	DioDigitalChannel *m_digitalChannel;
+
+	QString m_channelName;
+	QString m_channelType; // output or input
+
+	std::string m_iioAttrAvailableTypes; // iio attribute
```
adisuciu, line 62: Do you need to use std::string ? For consistency I'd use QString as it plays nicer with Qt classes.
Same for std::vector -> QList<> or even QStringList

plugins/swiotrefactor/res/resources_rc.py
```diff
@@ -0,0 +1,5115 @@
+# -*- coding: utf-8 -*-
+
+# Resource object code
+#
+# Created by: The Resource Compiler for PyQt5 (Qt v5.15.2)
+#
+# WARNING! All changes made in this file will be lost!
+
+from PyQt5 import QtCore
```
adisuciu, line 9: Is this file necessary ?

adisuciu (changes requested):
Looks pretty good

plugins/swiotrefactor/src/max14906/diodigitalchannelcontroller.cpp
```diff
@@ -0,0 +1,243 @@
...
+void DioDigitalChannelController::readCurrentLimitCmdFinished(Command *cmd)
+{
+	IioChannelAttributeRead *tcmd = dynamic_cast<IioChannelAttributeRead *>(cmd);
+	if(!tcmd) {
+		return;
+	}
+	if(tcmd->getReturnCode() >= 0) {
+		char *result = tcmd->getResult();
+		m_digitalChannel->m_currentLimitsCombo->combo()->setCurrentIndex(
+			m_digitalChannel->m_currentLimitsCombo->combo()->findText(result));
```
AlexandraTrifan, line 191: Should we validate the result of findText? If the result from IIO is an unexpected value, the find function will return -1 and the current index settings might have weird behaviour.

PR #1611, opened 2024-04-23
bufferpreviewer: fixed rolling mode bug

IonutMuthi:
data limit now has min and max value instead of beeing the absolte diff between two values
 use lower and upper axis bound instead of axis min and max value on drag to account for zoom level

### DIFF
--- a/gui/src/buffer_previewer.cpp
+++ b/gui/src/buffer_previewer.cpp
@@ -50,9 +50,11 @@ BufferPreviewer::BufferPreviewer(int pixelsPerPeriod, double wavePhase, QWidget 
 	, m_leftGateWidth(0)
 	, m_rightGateWidth(0)
 	, m_cursorVisible(true)
-{}
+{
+	m_wavePoints = new QPointF();
+}
 
-BufferPreviewer::~BufferPreviewer() { delete[] m_wavePoints; }
+BufferPreviewer::~BufferPreviewer() { delete m_wavePoints; }
 
 double BufferPreviewer::waveformPos() const { return m_waveformPos; }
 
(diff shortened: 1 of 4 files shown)


AlexandraTrifan (approved):

gui/src/widgets/plotbufferpreviewer.cpp
```diff
@@ -34,46 +35,54 @@ void PlotBufferPreviewer::setupBufferPreviewer()
 	connect(m_bufferPreviewer, &BufferPreviewer::bufferStopDrag, this, [=]() {});
 
 	connect(m_bufferPreviewer, &BufferPreviewer::bufferStartDrag, this, [=]() {
-		m_bufferPrevInitMin = m_plot->xAxis()->min();
-		m_bufferPrevInitMax = m_plot->xAxis()->max();
+		// reset the buffer preview position to current visible section
+		// using lower and upper bound to also consider zoom level
+		m_bufferPrevInitMin = m_plot->plot()->axisScaleDiv(m_plot->xAxis()->axisId()).lowerBound();
+		m_bufferPrevInitMax = m_plot->plot()->axisScaleDiv(m_plot->xAxis()->axisId()).upperBound();
```
andrei47w, line 41: replace axisScaleDiv with PlotAxis::visibleMin / PlotAxis::visibleMax

gui/src/buffer_previewer.cpp
```diff
@@ -52,7 +52,7 @@ BufferPreviewer::BufferPreviewer(int pixelsPerPeriod, double wavePhase, QWidget
 	, m_cursorVisible(true)
 {}
 
-BufferPreviewer::~BufferPreviewer() { delete[] m_wavePoints; }
+BufferPreviewer::~BufferPreviewer() { delete m_wavePoints; }
```
andrei47w, line 55: This alone doesn't fix the crash. m_wavePoints needs to be initialized in constructor
m_wavePoints(new QPointF)

andrei47w (changes requested):

andrei47w (approved):

PR #1614, opened 2024-04-23
core/emuwidget: Check iio-emu in system.

andreidanila1:


core/src/emuwidget.cpp
```diff
@@ -44,7 +44,7 @@ void EmuWidget::init()
 	Preferences *p = Preferences::GetInstance();
 	p->init("iio_emu_path", QCoreApplication::applicationDirPath());
 
-	m_emuPath = findEmuPath();
+	m_emuPath = (startIioEmuProcess("iio-emu")) ? "iio-emu" : findEmuPath();
```
adisuciu, line 47: Can you use if/else for both these statements instead of conditionals ? It makes the code a little hard to follow.

adisuciu (changes requested):

PR #1624, opened 2024-05-13
Bug regmap offset values

IonutMuthi:
Fixed 2 bugs on regmap 

* Map was only creating for AXI devices

* offset value and bit offset values were treated as hexa values

image (4)
image (3)
image (2)
image (5)


AlexandraTrifan:
Should this register disappear from the AD74412.xml file: https://github.com/analogdevicesinc/scopy/blob/6440bb9866aa37980898e8cc15f49eedd825fedd/plugins/regmap/xmls/AD74412.xml#L1929-L1950 ?

DIN_COUNTER register should only be available for AD74413R, when DIN_UPGRADE is true. There is no DIN_COUNTER in AD74412.

adisuciu (approved):

AlexandraTrifan (approved):

PR #1634, opened 2024-05-20
dev: remove plugins from build folder if they are disabled

IonutMuthi:
If a plugin that was enabled becomes disabled remove the ".so" file of that plugin


plugins/CMakeLists.txt
```diff
@@ -24,6 +24,14 @@ if(ENABLE_PLUGIN_TEST)
 	list(APPEND PLUGINS ${PLUGIN_NAME})
 endif()
 
+# If in DEV_MODE remove all ".so" files on each build to ensure olny ENABLED plugins have the ".so" file generated
+if(SCOPY_DEV_MODE)
+	file(GLOB SO_FILES "${SCOPY_PLUGIN_BUILD_PATH}/*.so")
```
adisuciu, line 29: Also delete *.dll if developing on windows ?

adisuciu (approved):

PR #1661, opened 2024-07-18
Add option for selecting the debugger version

Andrei-Fabian-Pop:
The preferences page now contains a checkbox that when selected (true) will only load the IIODebug (V2) and when it is unchecked (false) it will only load the Debugger (V1). The default is V2.


adisuciu:
I think you should merge the plugins to a single plugin with 2 instruments and the preference should be pluginspecific.

adisuciu (changes requested):
Merge instruments to the same plugin

andreidanila1 (approved):

adisuciu (approved):

PR #1664, opened 2024-07-19
IIOWidget: Add optional lazy loading

Andrei-Fabian-Pop:
The IIOWidget now has the ability to lazy load. By default (if not changed from the preference page) the IIOWidgets will only initialise on the first show event, not in the constructor, thus improving the loading time when connecting to a device. Deselecting this option will make the IIOWidgets initialise when created.


andreidanila1 (approved):

andreidanila1:
I tested it and it works fine. I think it would be a good idea to squash the commits because they all have the same purpose.

bindea-cristian (approved):

PR #1665, opened 2024-07-24
generic: plot widget print v0.1

IonutMuthi:


gui/src/plotwidget.cpp
```diff
@@ -259,6 +264,33 @@ void PlotWidget::setUnitsVisible(bool visible)
 	}
 }
 
+void PlotWidget::printPlot()
+{
+	// select file where to save
+	QString selectedFilter;
+	QString fileName =
```
adisuciu, line 271: Getting a file name should be separate from printing.
refactor this to printToFile(QString filename)

It should be possible to a QPixmap (or something similar) and have that be rendered in a file

adisuciu (changes requested):

adisuciu (approved):

PR #1677, opened 2024-08-14
emu: add configurations for emulator

IonutMuthi:


andreidanila1 (approved):

core/src/emuwidget.cpp
```diff
@@ -141,10 +157,13 @@ QStringList EmuWidget::createArgList()
 {
 	QString option = m_demoOptCb->currentText();
 	QStringList arguments;
-	arguments.append(option);
-	if(option.compare("generic") == 0) {
+
+	if(option.compare("adalm2000") != 0) {
```
adisuciu, line 161: This option could be part of the xml as well .. no ?

adisuciu (changes requested):

adisuciu (approved):

PR #1686, opened 2024-08-27
iio-widgets: Add conversion functions to IIOWidgets

Andrei-Fabian-Pop:


AlexandraTrifan:
can we check if iio widgets are for readonly attributes? if not, should we force the user of iio widgets to provide conversion functions for both directions?

AlexandraTrifan (approved):

PR #1719, opened 2024-09-20
iio-widgets: Use MenuSpinbox for RangeAttrUi strategy

Andrei-Fabian-Pop:


gui/src/widgets/menuspinbox.cpp
```diff
@@ -251,6 +273,12 @@ void MenuSpinbox::userInput(QString s)
 
 void MenuSpinbox::populateWidgets()
 {
+	// TODO: Review this function
+	if(!m_scalingEnabled) {
+		m_edit->setText(QString::number(m_value));
```
AlexandraTrifan, line 278: do we need to setup signal blockers before modifying the value of the edit box?

Andrei-Fabian-Pop: We might need to, good catch :sweat_smile:

adisuciu (approved):

PR #1738, opened 2024-09-30
Ci fix plugin path

bindea-cristian:


AlexandraTrifan (changes requested):
still need to add to .iss

AlexandraTrifan (approved):

PR #1753, opened 2024-10-10
gui/ConnectionLoadingBar: Add loading bar when connecting to devices

Andrei-Fabian-Pop:


AlexandraTrifan:
I think we should not merge this yet. If you unplug the board while connecting it will crash due to processEvents() function calls. We should investigate this further.

andrei47w (approved):

PR #1767, opened 2024-10-16
core: open external links from the About page.

AlexandraTrifan:


adisuciu:
void	setOpenExternalLinks(bool open) ? more explicit ? https://doc.qt.io/qt-5/qtextbrowser.html#openExternalLinks-prop ?

PR #1805, opened 2024-11-01
regmap: style updates

IonutMuthi:


plugins/regmap/src/regmapstylehelper.cpp
```diff
@@ -288,7 +288,7 @@ void RegmapStyleHelper::BitFieldDetailedWidgetStyle(BitFieldDetailedWidget *widg
 						}
                         )css");
 
-	style.replace("&&frameBackground&&", Style::getAttribute(json::theme::background_primary));
+	style.replace("&&frameBackground&&", Style::getAttribute(json::theme::interactive_subtle_disabled));
```
Andrei-Fabian-Pop, line 291: This style string might be better of in its own .qss file.

AlexandraTrifan (approved):

PR #1819, opened 2024-11-11
tools/style_helper: Fix generating style properties and build directory files

JJuanill:
Adjusted style_helper.py tool to generate style_properties.h

### DIFF
--- a/tools/style_generator.py
+++ b/tools/style_generator.py
@@ -34,9 +34,7 @@ def create_namespace_structure(extension: str, path: str, root_namespace: str = 
 
 # Replaces a placeholder property in a file with a given value
 def replace_property(in_file: str, value: str) -> bool:
-    out_file = os.path.join(os.path.join(build_folder, "style"), value + ".qss").replace("\\", "/")
-    out_dir = os.path.dirname(out_file)
-    os.makedirs(out_dir, exist_ok=True)
+    out_file = os.path.join(build_qss_folder, value + ".qss")
     search_text = "&&property&&"
 
     with open(in_file, 'r') as file:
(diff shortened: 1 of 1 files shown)


andrei47w:
After this modification the .qss files are being copied to build_folder/style instead of build_folder/style/qss and Scopy is not loading the style correctly. 
What is the problem you are trying to fix? We didn't have issues on any platform.

JJuanill:
The issue I am facing is when trying to build the project, it does not have a reference to `style::properties::<element>` (e.g. `style::properties::button::spinBox`). Looking at what is generated at style_properties.h it just shows two namespaces style and properties and nothing follows.

``` C++
// Do not modify this file! This was auto generated.
#ifndef STYLE_PROPERTIES_H
#define STYLE_PROPERTIES_H
#include <scopy-gui_export.h>

namespace style {
namespace properties {
} // namespace properties
} // namespace style


#endif // STYLE_PROPERTIES_H
```

Is this an intended behavior or is it supposed to have content?

JJuanill:
Looking at what is outputted at /styles/qss/, I pushed a fix that prevents the program from loading the styles. I think this has to do with how Windows handle local file locations. The original code captures forward slash '/' but checking what was outputted was a backslash '\' so I added the appropriate handles.

andrei47w:
For now I don't see a reason to merge these fixes since build on all platform work as expected (including windows). You may still use them for your local builds.
We are using mingw to build on windows and if you are using something different that might be why you are having issues.

gvanem:
I had the same issue with an empty `style_properties.h`. The patch from @JJuanill works fine for me.
I use clang-cl to build Scopy and it works fine. A real pity IMHO that MinGW64 is needed to built Scopy now.

The reason it's not noted in Windows-CI, is that the Python version used there is from the Msys2 project
(which do not seems to use `\` for `os.path.*()` functions).

PR #1872, opened 2025-01-10
docs: add default structure for all test cases

AlexandraTrifan:
**Below are some recommendations for formatting documentation files so that we all have a similar template.**

## User Guide
- Use the `.. note::` rst label.
- The information within the notes should follow the format: 
  - **User guide: :ref:`My plugin user guide<doc_reference>`**.

## Setup
- Name: **Setup environment**
- Use the same heading level as for the tests.
- Each setup should have a label so that it can be referenced from the tests.
- The names of the setups should be bold.
- The steps within the setup should be written as a bulleted list.

## Test Labels
- All test labels should be bold (UID, Description, Preconditions...). Also the **Expected result:** label should be bold.
- Expected result
  - If there is only one expected result, it must be written inline, otherwise all expected results must be written as a bulleted list.
- The UID and Description should be written inline with the labels, without bullets.
- Each test must contains the **Result** label
  - **Result:** PASS/FAIL 

## Preconditions
- Use a bulleted list.
- References to setups.
- OS label 
  - use capital letters for the label name 
  - options: ANY/ALL/a, list, of, operating, systems
    - Windows, Linux-x86_64, Linux-arm64, Linux-arm32, macOS 

## Steps
- Use a numbered list.
- Use a bullet for the **Expected result**.

## Prerequisites and Dependencies
- If they are generic, they can be written under the Setup environment section.
- If they are not generic, they can be written individually, for each test.

## Highlight field names
- Use bold for button names, field names, etc.

---
**NOTES**

- All the test cases PRs should be merged here first. 
- When all of them are ready to go, we can get this branch to main.
- [x] todo: change default doc branch to main instead of dev

---


docs/tests/general/core/dev_tests.rst
```diff
@@ -0,0 +1,592 @@
...
+    - Use :ref:`M2k.Emu <m2k-emu-dev>` setup.
+    - OS: Windows, Linux-x86_64, Linux-arm64, Linux-arm32, macOS.
+
+**Steps:**
+    1. Click on the device icon from device browser.
+    2. Click on **Connect** button.
+    3. Click on "Oscilloscope" tool.
+        - **Expected result:** The "Oscilloscope" tool is displayed and the 
+          tool menu item is highlighted.
+    4. Click on "Home" menu entry.
```
andreidanila1, line 245: Actual result missing.

andreidanila1 (approved):

PR #1878, opened 2025-01-17
plugin: Add ADMT Plugin for ADMT4000

JJuanill:
Adding implementation of ADMT4000 Harmonic Calibration plugin to Scopy


andrei47w:
I see the PR is still in progess, but I noticed a few things that might need changing before this is ready for complete review.

- create an xml for emulator so we can test the plugin without needing the hardware (https://github.com/analogdevicesinc/iio-emu/tree/main/tools)
- format code using clang-format version 10 (run tools/format.sh)
- revert change in the main CMakeLists.txt
- admt/CMakeLists:
	- remove unused liked libs from cmake (scopy-gr-util, scopy-iioutil, scopy-iio-widgets)
	- remove UI_LIST uses
- all larger stylesheets from inside .cpp files should be moved to qss using the style infrastructure
- migrate admtstylehelper.cpp to qss style and use Style::getChannelColor instead of redefining colors
- HorizontalSpinBox::setEnabled argument should be bool ? 
- rebase with main

plugins/admt/src/harmoniccalibration.cpp
```diff
@@ -0,0 +1,4766 @@
...
+void HarmonicCalibration::startAcquisition()
+{
+	isStartAcquisition = true;
+	acquisitionXPlotAxis->setInterval(0, acquisitionDisplayLength);
+
+	m_acquisitionDataThread =
+		QtConcurrent::run(this, &HarmonicCalibration::getAcquisitionSamples, acquisitionSampleRate);
+	m_acquisitionDataWatcher.setFuture(m_acquisitionDataThread);
+	m_acquisitionGraphThread =
+		QtConcurrent::run(this, &HarmonicCalibration::acquisitionPlotTask, acquisitionGraphSampleRate);
```
andrei47w, line 2581: This causes gui updates from outside the main thread and may cause unexpected behavior. You may use signals and slots when updating the gui. Other operations can be done in a different thread.

This causes a warning: QCoreApplication::sendPostedEvents: Cannot send posted events for objects in another thread

plugins/admt/src/harmoniccalibration.cpp
```diff
@@ -0,0 +1,4766 @@
...
+		QThread::msleep(sampleRate);
+	}
+}
+
+void HarmonicCalibration::startAcquisitionUITask()
+{
+	if(!m_acquisitionUIThread.isRunning()) {
+		isAcquisitionTab = true;
+		m_acquisitionUIThread =
+			QtConcurrent::run(this, &HarmonicCalibration::acquisitionUITask, acquisitionUITimerRate);
```
andrei47w, line 2770: This also causes gui updates from outside the main thread

plugins/admt/src/harmoniccalibration.cpp
```diff
@@ -0,0 +1,4766 @@
...
+}
+#pragma endregion
+
+#pragma region Calibration Methods
+void HarmonicCalibration::startCalibrationUITask()
+{
+	if(!m_calibrationUIThread.isRunning()) {
+		isCalibrationTab = true;
+		m_calibrationUIThread =
+			QtConcurrent::run(this, &HarmonicCalibration::calibrationUITask, calibrationUITimerRate);
```
andrei47w, line 2910: This also causes gui updates from outside the main thread

plugins/admt/src/harmoniccalibration.cpp
```diff
@@ -0,0 +1,4766 @@
...
+	else if(isCalibrationTab)
+		updateLineEditValue(calibrationMotorCurrentPositionLineEdit, current_pos);
+}
+
+bool HarmonicCalibration::resetGENERAL()
+{
+	bool success = false;
+
+	uint32_t resetValue = 0x0000;
+	if(deviceRegisterMap.at("ASIL ID") == "ASIL QM") {
```
andrei47w, line 2482: When using the emulator, this crashes Scopy since deviceRegisterMap is empty. If registers cannot be read it can give out an error, but it should not crash the app. Make sure to handle these edge cases.

plugins/admt/src/harmoniccalibration.cpp
```diff
@@ -0,0 +1,4766 @@
...
+
+	QCheckBox *cosineCheckBox = new QCheckBox("Cosine", acquisitionGraphChannelWidget);
+	ADMTStyleHelper::ColoredSquareCheckbox(cosineCheckBox, cosineColor);
+	connectCheckBoxToAcquisitionGraph(cosineCheckBox, acquisitionCosinePlotChannel, COSINE);
+
+	QCheckBox *radiusCheckBox = new QCheckBox("Radius", acquisitionGraphChannelWidget);
+	ADMTStyleHelper::ColoredSquareCheckbox(radiusCheckBox, channel5Pen.color());
+	connectCheckBoxToAcquisitionGraph(radiusCheckBox, acquisitionRadiusPlotChannel, RADIUS);
+
+	if(GENERALRegisterMap.at("Sequence Type") == 0) // Sequence Mode 1
```
andrei47w, line 478: This also crashes the app when using the emulator since GENERALRegisterMap is empty. For example you can replace these maps with QMap, as they will not cause a crash but you may still need to do some validations.

plugins/admt/xmls/admt4000.xml
```diff
@@ -0,0 +1,116 @@
+<?xml version="1.0" encoding="utf-8"?>
```
andrei47w, line 1: Move this file to plugins/admt/res/emu_xml/admt4000.xml. This way it should automatically be copied to plugins/emu_xml in the build folder

plugins/admt/test/tst_pluginloader.cpp
andrei47w, line 1: Fix the failing tests. You can see the output from github here:
https://github.com/analogdevicesinc/scopy/actions/runs/13852996380/job/38763773847?pr=1878#step:5:973

andrei47w (changes requested):

plugins/admt/src/harmoniccalibration.cpp
```diff
@@ -0,0 +1,5234 @@
...
+			resetCalculatedCoeffHex();
+		}
+	}
+}
+
+void HarmonicCalibration::calibrationLogWrite(QString message) { logsPlainTextEdit->appendPlainText(message); }
+
+void HarmonicCalibration::importCalibrationData()
+{
+	QString fileName = QFileDialog::getOpenFileName(this, tr("Import"), "",
```
andrei47w, line 4058: Use the general_use_native_dialogs preference, since on some platform native popups may not work. Here is an example:
https://github.com/analogdevicesinc/scopy/blob/ca5e34f625a70a9106889c72d0d6cc56ec75cdad/core/src/emuwidget.cpp#L383

plugins/admt/src/harmoniccalibration.cpp
```diff
@@ -0,0 +1,5234 @@
...
+void HarmonicCalibration::extractCalibrationData()
+{
+	QStringList filter;
+	filter += QString(tr("Comma-separated values files (*.csv)"));
+	filter += QString(tr("Tab-delimited values files (*.txt)"));
+	filter += QString(tr("All Files(*)"));
+
+	QString selectedFilter = filter[0];
+
+	QString fileName = QFileDialog::getSaveFileName(this, tr("Export"), "", filter.join(";;"), &selectedFilter,
```
andrei47w, line 4095: Same here, use general_use_native_dialogs

andrei47w (changes requested):

adisuciu (approved):

adisuciu:
This looks good .. 

Can you clean up the commit history a little bit ? 
We can also squash and have a single commit for this pull request.

PR #1889, opened 2025-01-29
Style script fix platform dependent path separator

andrei47w:
This should fix the style generation script on windows for users.
We did not encounter this issue since we are using MSYS2 to build on windows.


andrei47w:
This is a complete fix for https://github.com/analogdevicesinc/scopy/pull/1819
We've decided that this is an important fix since users may be using different environments to build Scopy.
Let me know if this works for you @JJuanill @gvanem

PR #1931, opened 2025-02-20
regmap: fix register dump and find file

IonutMuthi:


AlexandraTrifan:
still having some issues when writing the value of a register that's been dumped

AlexandraTrifan (approved):
looks good

PR #1948, opened 2025-02-26
m2k: signal generator infinite loop bug fix

IonutMuthi:
For small sps values on math channel the function "get_samples_count" will get stuck in an infinite loop due to size value reaching 0 before attempting "size <<=1 "

### DIFF
--- a/plugins/m2k/src/old/signal_generator.cpp
+++ b/plugins/m2k/src/old/signal_generator.cpp
@@ -2272,6 +2272,12 @@ size_t SignalGenerator::get_samples_count(unsigned int chnIdx, double rate, bool
 	}
 
 out_cleanup:
+
+	// for size 0 return 0 going further with this value will get stuck in an infinite loop
+	if(size == 0) {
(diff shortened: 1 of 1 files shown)


plugins/m2k/src/old/signal_generator.cpp
```diff
@@ -2272,6 +2272,12 @@ size_t SignalGenerator::get_samples_count(unsigned int chnIdx, double rate, bool
 	}
 
 out_cleanup:
+
+	// for size 0 return 0 going further with this value will get stuck in an infinite loop
+	if(size == 0) {
+		return 0;
```
AlexandraTrifan, line 2278: I think we should disable the Run button in this case.
What happens here is the the SigGen does not output anything, but the Run button is checked, suggesting that data is in fact generated on the DAC, which is not the case..

AlexandraTrifan (changes requested):

AlexandraTrifan (approved):

PR #1960, opened 2025-03-06
Preferences open preferences.ini button fix

andrei47w:


AlexandraTrifan (approved):
Tested on Pi, ubuntu and MacOS and everything seems to work as expected.

PR #2029, opened 2025-05-16
datalogger: added options to override scale and offset for channels

IonutMuthi:
added options to override scale and offset for data logger channels

### DIFF
--- a/plugins/datalogger/src/datamonitor/datamonitormodel.cpp
+++ b/plugins/datalogger/src/datamonitor/datamonitormodel.cpp
@@ -29,10 +29,22 @@
 using namespace scopy;
 using namespace datamonitor;
 
-DataMonitorModel::DataMonitorModel(QObject *parent) { setDataStorageSize(); }
+DataMonitorModel::DataMonitorModel(QObject *parent, double defaultScale, double defaultOffset)
+	: m_defaultScale(defaultScale)
+	, m_defaultOffset(defaultOffset)
+	, m_scale(defaultScale)
+	, m_offset(defaultOffset)
+{
+	setDataStorageSize();
+}
 
-DataMonitorModel::DataMonitorModel(QString name, QColor color, UnitOfMeasurement *unitOfMeasure, QObject *parent)
+DataMonitorModel::DataMonitorModel(QString name, QColor color, UnitOfMeasurement *unitOfMeasure, double defaultScale,
+				   double defaultOffset, QObject *parent)
 	: color(color)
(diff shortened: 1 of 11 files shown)


plugins/datalogger/src/dmm.cpp
```diff
@@ -66,14 +71,16 @@ QList<DmmDataMonitorModel *> DMM::getDmmMonitors(iio_context *ctx)
 						dmmInfo = m_hwmonDevices.value(static_cast<hwmon_chan_type>(type));
 					}
 					unitOfMeasurement = new UnitOfMeasurement(dmmInfo.name, dmmInfo.symbol);
-					dmmReadStrategy->setUmScale(dmmInfo.scale);
+					scale = scale * dmmInfo.scale;
```
AlexandraTrifan, line 74: I think we need to keep these 2 separated if we want to override scale. The user should only override the hardware scale, not the UM scale.

plugins/datalogger/src/datamonitor/datamonitormodel.cpp
```diff
@@ -200,11 +222,40 @@ void DataMonitorModel::addValue(double time, double value)
 		setMaxValue(value);
 	}
 
+	// Apply scaling, offset, and unit of measurement scale
+	double adjustedValue = (value + m_offset) * m_scale;
```
AlexandraTrifan, line 226: Will the overriden offset be provided in counts as well? If the new offset is provided as UM (V, Celsius, etc) it should be applied after scaling, otherwise it should be applied before before.

plugins/datalogger/src/datamonitor/datamonitormodel.cpp
```diff
@@ -200,11 +222,40 @@ void DataMonitorModel::addValue(double time, double value)
...
+}
+
+double DataMonitorModel::scale() const { return m_scale; }
+
+void DataMonitorModel::setScale(double newScale)
+{
+	double oldScale = m_scale;
+	m_scale = newScale;
+	for(int i = 0; i < ydata.size(); ++i) {
+		ydata[i] = ydata[i] / oldScale * newScale;
```
AlexandraTrifan, line 256: As mentioned below, we should keep scale and umScale separated.

AlexandraTrifan (changes requested):
Some changes are requested on "scale" separation. 
Also we need to clarify how the overriden values are provided.

AlexandraTrifan (approved):

PR #2035, opened 2025-05-23
ADC: fft averaging and other updates

andrei47w:


gui/include/gui/widgets/menucontrolbutton.h
```diff
@@ -107,12 +109,15 @@ class SCOPY_GUI_EXPORT CollapsableMenuControlButton : public QWidget, public Com
 	int count();
 	MenuControlButton *getControlBtn();
 	MenuControlButton *header() override { return m_ctrl; }
+	SmallOnOffSwitch *onOffSwitch();
+	SmallOnOffSwitch *enableOnOffSwitch(bool en);
```
AlexandraTrifan, line 113: should this have a void return type?

AlexandraTrifan:
Should we display somewhere on the plot status some info on whether averaging is enabled and size?

PR #2057, opened 2025-07-08
Pluto pkg

andreidanila1:


packages/pluto/plugins/pluto/CMakeLists.txt
```diff
@@ -26,7 +26,7 @@ message(STATUS "building plugin: " ${SCOPY_MODULE})
 
 project(scopy-${SCOPY_MODULE} VERSION 0.1 LANGUAGES CXX)
 
-set(PLUGIN_DISPLAY_NAME AD936X)
+set(PLUGIN_DISPLAY_NAME PLUTO)
```
IonutMuthi, line 29: this should be AD936X not PLUTO since it covers all devices of ad936x family

PR #2081, opened 2025-08-26
Ad936x refactoring

IonutMuthi:


packages/ad936x/plugins/ad936x/src/fmcomms5/fmcomms5calibration.cpp
```diff
@@ -0,0 +1,632 @@
...
+
+		if(dev_name && QString(dev_name).compare("ad9361-phy-B", Qt::CaseInsensitive) == 0) {
+			secondDevice = dev;
+			foundDevices++;
+		}
+
+		if(foundDevices == 2) {
+			break;
+		}
+	}
```
dNechita: Why not directly search for the 2 devices?
```
iio_context_find_device(m_ctx, "ad9361-phy");
iio_context_find_device(m_ctx, "ad9361-phy-B");
```

packages/ad936x/plugins/ad936x/src/ad936x/ad936x.cpp
```diff
@@ -97,7 +97,7 @@ AD936X::AD936X(iio_context *ctx, QWidget *parent)
 	controlsLayout->addWidget(scrollArea);
 
 	if(m_ctx != nullptr) {
-
+		// this should work for any device from AD936x family
```
dNechita, line 100: Does it matter if the below for loop will first encounter "cf-ad936x-lpc" instead of "ad936x-phy"? If yes, then maybe try to check if the name starts with "ad936" instead of `contains()`.

PR #2112, opened 2025-11-03
iiowidgets: added range to ui conversion function

IonutMuthi:


iio-widgets/src/iiowidget.cpp
```diff
@@ -184,6 +187,18 @@ void IIOWidget::convertDStoUI(QString data, QString optionalData)
...
+
+	if(m_RangetoUI && !optionalData.isEmpty()) { // the range conversion is done with it's own function
+		QString cleanData = optionalData;
+		cleanData.remove('[').remove(']'); // Remove brackets
+		QStringList values = cleanData.split(" ", Qt::SkipEmptyParts);
+		QStringList converted;
+		for(const QString &value : values) {
+			converted.append(m_RangetoUI(value)); // Apply to each value
+		}
+		optionalData = converted.join(" ");
```
Andrei-Fabian-Pop, line 199: Shouldn't the `optionalData` also have the brackets? The optional data before the if clause looks like `[x y z]` and the one after looks like `x y z`.

PR #2147, opened 2025-12-16
CI and Dockers Update

bindea-cristian:
- updated dependencies
     - updated Gnuradio
     - deps installed using package managers
 - updated Ubuntu 20.04 Docker Image to 24.04 for x86_64 Appimage Workflow
 - updated Flatpak Docker Image with newer versions of libraries and Ubuntu 22.04
 - added Genalyzer
 - updated Msys version and Windows starting image from windows:ltsc2019 to windows:ltsc2022

### DIFF
--- a/ci/flatpak/Makefile
+++ b/ci/flatpak/Makefile
@@ -19,7 +19,7 @@ repo: build
 	flatpak build-export $@ $<
 
 build: preprocess
-	flatpak-builder --verbose --keep-build-dirs --ccache --user --disable-rofiles-fuse --arch=$(ARCH) $@ $(SRC)
+	flatpak-builder --verbose --keep-build-dirs --ccache --user --disable-rofiles-fuse --disable-updates --arch=$(ARCH) $@ $(SRC)
 
 preprocess: clean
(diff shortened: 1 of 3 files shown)


AlexandraTrifan (approved):

AlexandraTrifan (changes requested):
- windows portable build (probably the installer as well) do not properly load the m2k and imu plugin (maybe more, but only these for now)

AlexandraTrifan:
> I've fixed the Windows installer here [6fab326](https://github.com/analogdevicesinc/scopy/commit/6fab326894b1cb49fb9b776f5768f86f91ee32c3), there were some missing dlls.

Ran some tests on the Windows build - all plugins are correctly loaded

bindea-cristian:
Updating the starting Docker Image for the x86_64 AppImage Workflow from Ubuntu 20.04 to 24.04 is a issue for users with Ubuntu older that 24.04.
This update was done to remove possible vulnerabilities of old components, but due to missing GLIBC  2.38 in the Ubuntu 22, the AppImage will not start.
Currently searching for a solution...

PR #2190, opened 2026-02-16
Some bugfixes I found while working on the Android version of Scopy

bindea-cristian:
- some memory leaks
 - some dynamic_cast issues

### DIFF
--- a/gr-util/include/gr-util/grtopblock.h
+++ b/gr-util/include/gr-util/grtopblock.h
@@ -28,6 +28,7 @@
 
 #include <gnuradio/top_block.h>
 
+#include <QFuture>
 #include <QObject>
 
 namespace scopy::grutil {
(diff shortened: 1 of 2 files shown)


gui/src/buffer_previewer.cpp
andreidanila1, line 1: It might be a good idea to add a null check for m_wavePoints at the beginning of BufferPreviewer::drawBufferWave

gr-util/src/genalyzer_impl.cpp
```diff
@@ -190,13 +190,20 @@ GenalyzerConfig genalyzer_fft_vii_impl::get_config() const { return d_genalyzer_
...
+		free(d_analysis->rvalues);
+	}
 	d_analysis->results_size = 0;
 	d_analysis->rkeys = nullptr;
 	d_analysis->rvalues = nullptr;
 }
 
 void genalyzer_fft_vii_impl::set_analysis_results(size_t results_size, char **rkeys, double *rvalues)
 {
+	clear_analysis_results();
```
andreidanila1: This change causes a crash when trying to enable the genalyzer analysis.

PR #2192, opened 2026-02-17
ADC UI rework and minor bug fixes

andrei47w:


iioutil/src/iiopingtask.cpp
```diff
@@ -22,25 +22,68 @@
...
+	, m_disconnectBtn(nullptr)
+	, m_warningShown(false)
+{
+	setupWarningWidget();
+}
+
+IIOPingTask::~IIOPingTask()
+{
+	if(m_warningWidget) {
+		delete m_warningWidget;
```
AlexandraTrifan: can we give it a parent and let it cleanup?

core/src/deviceimpl.cpp
```diff
@@ -358,6 +366,25 @@ void DeviceImpl::onConnectionFailed()
 	disconnectDev();
 }
 
+void DeviceImpl::onConnectionLost(QWidget *warningWidget)
+{
+	if(!warningWidget) {
+		return;
+	}
+
+	bool confirmBeforeDisconnect = Preferences::get("general_confirm_on_connection_lost").toBool();
```
AlexandraTrifan: can preferences get fail? should we validate this?

andrei47w: if QVariant is invalid, toBool() returns false

PR #2223, opened 2026-03-17
Js tests infrastructure updates

IonutMuthi:


CMakeLists.txt
```diff
@@ -177,6 +177,12 @@ if(ENABLE_TESTING)
 	add_subdirectory(tests)
 endif()
 
+option(JS_TESTS_ENABLED "Copy JS test files to build directory" OFF)
```
bindea-cristian: You can use the already defined option: **ENABLE_SCOPYJS** from _pluginbase/CMakeLists.txt_.
I recommend moving the definition here, in the main CMake.
And you can leave it **ON** by default.

bindea-cristian (changes requested):

PR #2229, opened 2026-03-18
Add Debian package (.deb) build infrastructure for ARM AppImages

NicuS03:
Adds CI steps and packaging scripts to produce .deb packages from the ARM64/armhf AppImage builds, including debian control files, templates, configs, and a build script that repackages the AppImage into a proper Debian package.

### DIFF
--- a/.github/workflows/appimage-arm64.yml
+++ b/.github/workflows/appimage-arm64.yml
@@ -71,7 +71,7 @@ jobs:
             sudo apt-get install -y dpkg-dev debhelper devscripts rsync crossbuild-essential-arm64
             ./ci/packaging/scopy-deb-build.sh scopy64 $GITHUB_WORKSPACE/Scopy-arm64.AppImage
 
-        - uses: actions/upload-artifact@v4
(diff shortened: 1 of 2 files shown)


ci/packaging/debian/scopy.desktop.template
```diff
@@ -0,0 +1,10 @@
+[Desktop Entry]
+Version=1.0
```
AlexandraTrifan, line 2: should this be substituted by Scopy's version?

AlexandraTrifan:
Overall this looks good to me. 
I would have an addition to make, if possible.
In this PR: https://github.com/analogdevicesinc/scopy/pull/2230/changes we're adding policykit files for running Scopy as admin when ran from the Kuiper menu. These files should also be added to the debian package, from what I understand. Can we do that in this PR directly?

ci/packaging/debian/scopy.desktop.template
NicuS03: The version from the .desktop application os the Destkop Entry version.

PR #2230, opened 2026-03-18
homeaddpage: Add sudo restart dialog for local IIO devices.

andreidanila1:
It's a linux feature. If possible, please test it on Ubuntu 22, Ubuntu 24, and the latest Kuiper release. The video below shows how it should work.

[Screencast from 2026-03-18 16-50-39.webm](https://github.com/user-attachments/assets/9e7b306b-0f05-4d30-89c6-29cba03080e1)

### DIFF
--- a/ci/general/org.adi.scopy.policy
+++ b/ci/general/org.adi.scopy.policy
@@ -1,0 +1,22 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE policyconfig PUBLIC
+  "-//freedesktop//DTD PolicyKit Policy Configuration 1.0//EN"
+  "http://www.freedesktop.org/standards/PolicyKit/1/policyconfig.dtd">
+
+<policyconfig>
+  <vendor>Analog Devices Inc.</vendor>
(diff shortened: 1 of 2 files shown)


core/include/core/application_restarter.h
```diff
@@ -39,8 +39,13 @@ class SCOPY_CORE_EXPORT ApplicationRestarter
 
 	int restart(int exitCode);
 	static void triggerRestart();
+#ifdef __linux__
+	static void triggerRestartWithSudo();
+#endif
 
 private:
+	static QStringList buildSudoEnvArgs();
```
IonutMuthi: should this not be included only on linux?  on other platforms might generate warning

core/src/application_restarter.cpp
```diff
@@ -66,3 +71,50 @@ void ApplicationRestarter::triggerRestart()
...
+}
+
+#ifdef __linux__
+void ApplicationRestarter::triggerRestartWithSudo()
+{
+	if(getuid() == 0) {
+		return;
+	}
+
+	ApplicationRestarter *instance = GetInstance();
```
IonutMuthi: Check if instance is nullptr

PR #2242, opened 2026-04-08
tools: as scopy ss automatiosn

IonutMuthi:


bindea-cristian (approved):
It needs to be refined, there are a lot of duplicated images and screenshots that don't make any sense. But the majority of them are really useful when checking the UI.

The only problem is that when running the script locally, the first set of screenshots contain only the _What's new_ pop-up. Maybe the script should start with a restart of the application after setting the preferences.

As a first iteration of this feature it works well.

andreidanila1 (approved):

PR #2256, opened 2026-04-29
AD6676 package

IonutMuthi:
Adds AD6676 package from iio-oscilloscope to scopy


docs/tests/plugins/ad6676/ad6676_tests.rst
```diff
@@ -0,0 +1,479 @@
+.. _ad6676_tests:
+
+AD6676 Test Suite
+================================================================================
+
+**Prerequisites:**
+    - Scopy v2.1.0 or later with AD6676 plugin installed
```
AlexandraTrifan: Scopy 2.3.0 or later

js/testAutomations/ad6676/ad6676Tests.js
```diff
@@ -0,0 +1,440 @@
+/*
+ * Copyright (c) 2025 Analog Devices Inc.
```
AlexandraTrifan: same as previous PRs

js/testAutomations/ad6676/ad6676Tests.js
```diff
@@ -0,0 +1,440 @@
...
+// Source: docs/tests/plugins/ad6676/ad6676_tests.rst
+
+// Load test framework
+evaluateFile("../js/testAutomations/common/testFramework.js");
+
+// Test Suite: AD6676 Plugin Tests
+TestFramework.init("AD6676 Plugin Tests");
+
+// Connect to device
+if (!TestFramework.connectToDevice("ip:10.48.69.131")) {
```
AlexandraTrifan: could we pass the uri/ip as parameter and not have it hardcoded here?
or just put a placeholder?

IonutMuthi: this is supposed to be a placeholder and replaced while testing. This one is leftover from the testing and will be changed to "ip:127.0.0.0"

packages/ad6676/plugins/ad6676plugin/src/ad6676.cpp
```diff
@@ -0,0 +1,428 @@
...
+		return;
+
+	double freqMhz = adcFreqMhz.toDouble();
+	if(qFuzzyIsNull(freqMhz))
+		return;
+
+	auto *bwSpinbox = m_bwWidget->findChild<gui::MenuSpinbox *>();
+	if(bwSpinbox) {
+		bwSpinbox->setMinValue(0.005 * freqMhz); // 0.5% of ADC freq
+		bwSpinbox->setMaxValue(0.05 * freqMhz);	 // 5% of ADC freq
```
AlexandraTrifan: Not sure if this works properly or if I am testing it properly.
Having adc freq at 3200 MHz gives a max bandwidth of 160 MHz (which is correct and the spinbox is limited to it).
Having the ADC Freq at 2925 MHz (min value) should give a max bandwidth of around 146. But I am still able to set the value to 150 (or 160 for example)

maybe also update the range of the iio widget?

packages/ad6676/plugins/ad6676plugin/src/ad6676.cpp
```diff
@@ -0,0 +1,428 @@
...
+
+	QVBoxLayout *layout = new QVBoxLayout();
+	layout->setSpacing(10);
+	layout->setContentsMargins(5, 5, 5, 5);
+	section->setLayout(layout);
+
+	QLabel *label = new QLabel("Bandwidth Settings");
+	layout->addWidget(label);
+
+	// bandwidth: IIO in Hz, UI in MHz. Initial range 14.625–160 MHz (= 0.5%–5% of 2925 MHz).
```
AlexandraTrifan, line 285: initial value for the emu is 3200 MHz for freq; and 160 is 5% of 3200. this might be related to what i commented above?

packages/ad6676/plugins/ad6676plugin/src/ad6676plugin.cpp
```diff
@@ -0,0 +1,142 @@
...
+{
+	SCOPY_PLUGIN_ICON(":/gui/icons/adalm.svg");
+	return true;
+}
+
+void Ad6676Plugin::loadToolList()
+{
+	m_toolList.append(SCOPY_NEW_TOOLMENUENTRY("ad6676tool", "AD6676",
+						  ":/gui/icons/" + Style::getAttribute(json::theme::icon_theme_folder) +
+							  "/icons/gear_wheel.svg"));
```
AlexandraTrifan: wrong gear wheel icon

AlexandraTrifan (changes requested):

AlexandraTrifan (approved):

PR #2259, opened 2026-04-29
CN0357: package

IonutMuthi:
Adds CN0357 package from iio-oscilloscope to scopy

!! This was only tested on a reversed engineered emulated device created from iio-oscilloscope source code and should be tested with actual hardware !!


packages/cn0357/plugins/cn0357/include/cn0357/cn0357.h
```diff
@@ -0,0 +1,60 @@
+/*
+ * Copyright (c) 2025 Analog Devices Inc.
```
AlexandraTrifan: license header year can change

packages/cn0357/plugins/cn0357/src/cn0357.cpp
```diff
@@ -0,0 +1,117 @@
...
+{
+	SCOPY_PLUGIN_ICON(":/gui/icons/adalm.svg");
+	return true;
+}
+
+void Cn0357Plugin::loadToolList()
+{
+	m_toolList.append(SCOPY_NEW_TOOLMENUENTRY("cn0357tool", "CN0357",
+						  ":/gui/icons/" + Style::getAttribute(json::theme::icon_theme_folder) +
+							  "/icons/gear_wheel.svg"));
```
AlexandraTrifan: wrong wheel icon

AlexandraTrifan (approved):

PR #2274, opened 2026-06-30
ADC: Frequency markers crash fix

andrei47w:
this also includes device name in markers names to avoid maker labels being overwritten


IonutMuthi (approved):
I tested this on windows 11 and it worked with no issues

AlexandraTrifan (approved):

PR #2281, opened 2026-07-03
dds: various fixes for start/stop/resetting scales and frequencies.

AlexandraTrifan:
If a different application configured the DDS to run, Scopy should allow the user to select the non resetting startup and shutdown in order not to break the other configurations.

Changes to check or fix:

- [x] Using the on_connect/on_shutdown preference, Scopy should not reset the scales and "raw" enables of DDS channels. This can be tested by connecting the IIO-OSC and then connecting/disconnecting Scopy while the OSC generates a Tone.
- [x] Switching between DDS modes no longer resets the scale to -Inf, disabling the DDS channels. Again, this can be tested with the IIO OSC running in the background, while Scopy switches between modes just for configuration visualizing purposes, without messing up the initial config.
- [x] Scopy reads the config values when switching from Buffer/Disabled to DDS mode, such that it can detect if any Tone are configured and automatically switch the TX to that TONE.

### DIFF
--- a/packages/generic-plugins/plugins/dac/src/ddsdacaddon.cpp
+++ b/packages/generic-plugins/plugins/dac/src/ddsdacaddon.cpp
@@ -26,6 +26,7 @@
 #include "txtone.h"
 #include "dac_logging_categories.h"
 
+#include <pluginbase/preferences.h>
 #include <gui/widgets/menucontrolbutton.h>
 #include <gui/widgets/menucombo.h>
 #include <gui/mapstackedwidget.h>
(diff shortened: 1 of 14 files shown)


packages/generic-plugins/plugins/dac/src/ddsdacaddon.cpp
```diff
@@ -125,8 +125,12 @@ QWidget *DdsDacAddon::setupDdsTx(TxNode *txNode)
 		auto mode = cb->itemData(idx).toInt();
 		ddsModeStack->show(QString::number(mode));
 		TxMode *current = dynamic_cast<TxMode *>(ddsModeStack->get(QString::number(mode)));
+		if (mode == TxMode::DISABLED) {
```
andreidanila1: I think the null check of current should be moved here.

andreidanila1 (approved):

IonutMuthi (approved):

PR #2282, opened 2026-07-03
ADC: add global ch settings for each device

andrei47w:
- improved UI toggle all channels button from the device section 

before:
Screenshot from 2026-07-03 11-57-09

after:
Screenshot from 2026-07-03 13-32-58

device settings:
Screenshot from 2026-07-03 11-45-42

### DIFF
--- a/packages/generic-plugins/plugins/adc/src/freq/fftmenucontrols.cpp
+++ b/packages/generic-plugins/plugins/adc/src/freq/fftmenucontrols.cpp
@@ -1,0 +1,133 @@
+/*
+ * Copyright (c) 2026 Analog Devices Inc.
+ *
+ * This file is part of Scopy
+ * (see https://www.github.com/analogdevicesinc/scopy).
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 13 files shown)


packages/generic-plugins/plugins/adc/src/freq/fftmenucontrols.h
```diff
@@ -0,0 +1,67 @@
+/*
+ * Copyright (c) 2024 Analog Devices Inc.
```
andreidanila1: Copyright year

packages/generic-plugins/plugins/adc/src/freq/fftmenucontrols.cpp
```diff
@@ -0,0 +1,133 @@
+/*
+ * Copyright (c) 2024 Analog Devices Inc.
```
andreidanila1: Copyright year

andreidanila1 (approved):

AlexandraTrifan (approved):
all good, besides the fact that clicking the warning sign makes the menus look weird when Scopy is not full screen

Image

IonutMuthi (approved):

PR #2293, opened 2026-07-29
core: standardize on C++20 and raise CMake min version to 3.16

IonutMuthi:
The CMAKE_CXX_STANDARD 20 requires CMake >= 3.12
CMake 3.16 has a better compatibility with Qt6
The current CI use CMake 3.29
This change also fixed the warnings related to "=" being deprecated in CXX 20 standard


bindea-cristian (changes requested):
Make sure to also change the cmake_minimum_required version in the package generator templates too.

packages/generic-plugins/plugins/datalogger/src/menus/monitorselectionmenu.cpp
andreidanila1, line 107: Should "=" be removed?

andreidanila1 (approved):

bindea-cristian (approved):

PR #2296, opened 2026-08-12
ci: Fix find_package(Boost) for Ubuntu 26.04

Andrei-Fabian-Pop:
Boost.System is header-only since Boost 1.19 and its compiled library was removed in Boost 1.89, so it must not be requested as a component. This is especially visible in Ubuntu 26.04 since it is the latest apt version.

Using this fix, I managed to build Scopy (Qt5, tag: v2.2.1) on Ubuntu 26.04.


AlexandraTrifan:
https://github.com/analogdevicesinc/scopy/commit/46ec05c1db537169abc7d16cb7a29ed70b11e10c

I think this is no longer valid since we removed lots of dependencies on main

maybe this is something that can be applied on v2.2.x-maint branch which is the one used for a possible next future patch release

PR #2297, opened 2026-08-20
Scopy deps rework

IonutMuthi:


ci/arm/arm_cross_build_process.sh
```diff
@@ -220,89 +221,26 @@ build_libad9361() {
 	popd
 }
 
-build_libm2k() {
```
AlexandraTrifan: we need to bring this back, not drop libm2k 
we discussed whether libm2k is brought into the build by the m2k package
but the docker image can contain libm2k, or at least scripts to build it and just opt it out from this interm version

andreidanila1 (approved):

